PR #6, opened 2018-08-08
Mark two Google extensions as complete

antiagainst:


antiagainst:
No semantic change, so I'm not sure whether I should bump the version number here. Let me know if that's the convention.

PR #16, opened 2018-12-12
Update language for NonUniformEXT 

dgkoch:
to be consistent with the Uniform definition after Issue 317.

Issue: SPIR-V Khronos Internal #407


dgkoch:
cc @johnkslang

PR #24, opened 2019-04-12
Tidy up, rationalize, and update the set of extension files

johnkslang:
The commits are simple themes; it is much easier to see this PR one commit at a time.

- make all extensions have a {XXXX.asciidoc, XXXX.html} pair
- make all extensions up to date WRT Khronos internal copies

Note that different tool chains are used by different vendors, so this repository does not include a building mechanism for asciidoc -> html.


extensions/KHR/SPV_KHR_16bit_storage.asciidoc
```diff
@@ -0,0 +1,261 @@
...
+|1|2016-11-22|Alexander Galazin|*Initial revision*
+|2|2016-11-28|Alexander Galazin|*Address first round of feedback*
+|3|2016-12-01|Alexander Galazin|*Removed combined Load/Store and Convert instructions.
+Renamed capabilities and described them in terms of storage classes.*
+|4|2016-12-08|David Neto|*Assigned token numbers*
+|5|2016-12-14|Alexander Galazin| *Renamed the extension. Removed changes to the default rounding modes. Made StorageUniform16 dependent on StorageUniformBufferBlock16*
+|6|2017-02-22|JohnK| *Clarified that conversions for changing width can only change the width, not the fundamental type domain.*
+|7|2017-03-15|Alexander Galazin| *Clarified that FP Rounding mode can be used only if the capabilities from this extension are enabled*
+|8|2017-03-23|Alexander Galazin| Added interactions with *SPV_KHR_uniform_buffer_storage_class*
+|9|2018-06-11|Alexander Galazin| Added clarifications for SPIR-V issue 319
```
jeffbolznv: @johnkslang did you also check that the HTML files are up to date? https://www.khronos.org/registry/spir-v/extensions/KHR/SPV_KHR_16bit_storage.html is out of date (is missing revision 9), and I don't see it included in this change.

johnkslang: I did, and in fact 3 were out of date, which is what the 3rd commit fixed.

But, this branch and master both show that revision 9 is already in place.  I wonder about your link.  If I navigate to the normal "blob" type view from the top, for either master or tidy-up branches, I see the revision 9 change in the extension.

So, I'm thinking there is some cache problem with that non-blob link.

jeffbolznv: OK, I see that the github version is up to date. It is just the khronos.org mirror that is stale. Unfortunately, the khronos.org mirror is the first hit on Google, at least for extensions old enough to be present there.

The links at https://www.khronos.org/registry/spir-v/ all point to github now. Maybe the khronos.org mirror should just be deleted, and then google would stop linking to them? @outofcontrol do you control the mirror on khronos.org or know the history here?

PR #28, opened 2019-05-06
Publish Intel FPGA loop control extension

mkinsner:


extensions/INTEL/SPV_INTEL_fpga_loop_controls.asciidoc
```diff
@@ -0,0 +1,136 @@
...
+== Dependencies
+
+This extension is written against the SPIR-V Specification,
+Version 1.3 Revision 6.
+
+This extension requires SPIR-V 1.0.
+
+== Overview
+
+This extension adds a mechanism to allow more than the current limit of 32 loop controls.  It uses that new mechanism to add new loop controls for FPGA targets.
```
johnkslang: Because one of those bits could be used as a continuation bit to the next word, it's not quite accurate to say there is a limit of only 32 controls.

The new controls proposed do seem in good company with some existing controls, so I think Khronos should look at reserving some existing bits for this purpose.

mkinsner: Hi John,

I've updated the overview wording through an additional commit.

We don't disagree with your comments.  Are you asking for one of the following options up front, or just making an observation with your comment?

1. We can make some of these controls use bitfield locations in the future, once there is a way to allocate bits for specific extensions.  I'd expect a future revision of this extension to factor in that change.

2. We could split the extension into two parts, one that defines extended loop controls trailing as additional operands, and one that defines the additional loop controls.  We need the additional operands for some controls (bitfield isn't sufficient) because there can be repeated instances of certain controls on the same OpLoopMerge - e.g. DependencyArrayINTEL.  Our thought is that Khronos should define the 0x80000000 bitfield longer term (matching the extension functionality), and then we can remove that part from this extension.

I think two bitfield values need to be formally defined longer term: (1) what we've defined here as 0x80000000, where additional operands define loop controls which may repeat; (2) the extended bitfield case, indicating that the following word is an additional bitfield for loop controls.

Otherwise, please let me know what we should do in order for this to be merged.  Are you suggesting that we should switch to formal bitfield allocations for some of these controls (the non-repeating ones) before merging the extension to the registry?

johnkslang: Intel has the final decision of what goes in the registry within an Intel vendor-specific extension. I just wanted to leave this open long enough to first gather feedback and have some discussion. This can potentially avoid greater pain further down the road.

mkinsner: The extension has been updated to use the loop control bitfield allocation mechanism now in SPIRV-Headers/spir-v.xml.  Much cleaner now!

README.md
```diff
@@ -62,4 +62,5 @@ Khronos SPIR-V Registry](https://www.khronos.org/registry/spir-v/).
 1. [SPV_EXT_physical_storage_buffer         ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/EXT/SPV_EXT_physical_storage_buffer.html)
 1. [SPV_INTEL_fpga_memory_attributes        ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_fpga_memory_attributes.html)
 1. [SPV_NV_cooperative_matrix               ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/NV/SPV_NV_cooperative_matrix.html)
-1. [SPV_INTEL_shader_integer_functions2     ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_shader_integer_functions2.html)
\ No newline at end of file
+1. [SPV_INTEL_shader_integer_functions2     ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_shader_integer_functions2.html)
```
johnkslang: Is this diff due to line-ending differences?  I think we should keep it all as linux style, but don't know the actual state right now.

johnkslang: (Current repository state is linux style.)

mkinsner: I don't believe this is an OS line-ending style issue.  Unix utilities and git appear to complain is there isn't a trailing \n in a file, plus show the addition of it as a modification in diff.  I believe that the previous commit must have been missing a trailing \n.  vi appends one at end of file by default, but not all editors do.

I don't think there is anything to do here.  I confirmed that my added line (new end of file) has a trailing \n in the binary dump.

Git diff says that the old commit didn't have a trailing \n:

```
git diff 4b77d36f38dee2fa95ede1a13d5b0c8c094411ba README.md
diff --git a/README.md b/README.md
index fe1f59f..a4507ce 100644
--- a/README.md
+++ b/README.md
@@ -62,4 +62,5 @@ Khronos SPIR-V Registry](https://www.khronos.org/registry/spir-v/).
 1. [SPV_EXT_physical_storage_buffer         ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/EXT/SPV_EXT_physical_storage_buffer.html)
 1. [SPV_INTEL_fpga_memory_attributes        ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_fpga_memory_attributes.html)
 1. [SPV_NV_cooperative_matrix               ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/NV/SPV_NV_cooperative_matrix.html)
-1. [SPV_INTEL_shader_integer_functions2     ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_shader_integer_functions2.html)
\ No newline at end of file
+1. [SPV_INTEL_shader_integer_functions2     ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_shader_integer_functions2.html)
+1. [SPV_INTEL_fpga_loop_controls            ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/INTEL/SPV_INTEL_fpga_loop_controls.html)
```

https://thoughtbot.com/blog/no-newline-at-end-of-file

PR #30, opened 2019-05-20
SPV_AMD_shader_ballot: Extension spec update

DominikWitczakAMD:
Fixed an issue in the section describing how to use the new functionality. Fixed MbcntAMD's return type.


johnkslang:
Do you want to fix the white space so that history shows a relevant diff?  Or, just go with it as is?

DominikWitczakAMD:
Apologies for the late reply. Let's just go with what we have.

PR #32, opened 2019-06-05
Update memory model extension wording to read better against 1.4

jeffbolznv:


johnkslang:
Thanks Jeff. I didn't intend to request that we have to put the tag-linking in the extension document, though it certainly makes the next steps faster.

PR #39, opened 2019-06-13
Add Volatile memory semantics bit

jeffbolznv:
Add a new "VolatileKHR" memory semantics bit to SPV_KHR_vulkan_memory_model.

### DIFF
--- a/extensions/KHR/SPV_KHR_vulkan_memory_model.asciidoc
+++ b/extensions/KHR/SPV_KHR_vulkan_memory_model.asciidoc
@@ -30,8 +30,8 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2019-02-19
-| Revision           | 3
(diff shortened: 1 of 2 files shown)


extensions/KHR/SPV_KHR_vulkan_memory_model.asciidoc
```diff
@@ -30,7 +30,7 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2019-02-19
+| Last Modified Date | 2019-06-13
 | Revision           | 3
```
dneto0: Should probably be revision 4.  :-)

dneto0: Should be revision 4

jeffbolznv: fixed

PR #46, opened 2019-07-17
Initial revision of Intel blocking pipes SPIR-V extension

mkinsner:


extensions/INTEL/SPV_INTEL_blocking_pipes.asciidoc
```diff
@@ -0,0 +1,176 @@
...
+_Packet Size_ and _Packet Alignment_ must satisfy the following: +
+- 1 &lt;= _Packet Alignment_ &lt;= _Packet Size_. +
+- _Packet Alignment_ must evenly divide _Packet Size_. +
+
+For concrete types, _Packet Alignment_ should equal _Packet Size_.  For aggregate types, _Packet Alignment_ should be the size of the largest primitive type in the hierarchy of types.
+| Capability: +
+*BlockingPipesINTEL*
+
+| 5 | 5946 | <id> +
+_Pipe_ | <id> +
```
AlexeySotkin: Should it have the result type and result id operands, like blocking instructions have?

GarveyJoe: They aren't needed because these blocking variants don't return anything.  They block until they succeed.

AlexeySotkin: Ok, thanks.
@mrSidims please update https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/255 as it seems to generate those operands.

PR #47, opened 2019-07-30
Add SPV_GOOGLE_user_type.

ehsannas:


extensions/GOOGLE/SPV_GOOGLE_user_type.asciidoc
```diff
@@ -0,0 +1,178 @@
...
+
+Dependencies
+------------
+
+This extension is written against the SPIR-V Specification,
+Version 1.2 Revision 4, Unified.
+
+This extension requires SPIR-V 1.0.
+
+This extension also subsumes the SPV_GOOGLE_decorate_string extension.
```
johnkslang: This sounds like this is a superset of SPV_GOOGLE_decorate_string, the latter no longer being needed.

But, do you mean it just includes it, so that it does not have to be independently declared?

> Use of this extension implicitly includes SPV_GOOGLE_decorate_string, with no declaration needed for SPV_GOOGLE_decorate_string.

That is when using SPV_GOOGLE_user_type, if one uses a feature of SPV_GOOGLE_decorate_string are they still using SPV_GOOGLE_decorate_string (minus declaration) or are they only using SPV_GOOGLE_user_type?

johnkslang:
I have one question to clarify in the overview.

Note though this is not blocking.  Once a vendor-specific extension is posted here, it is available for use.

extensions/GOOGLE/SPV_GOOGLE_user_type.asciidoc
ehsannas: > do you mean it just includes it, so that it does not have to be independently declared?

Yes, that's my understanding. Similar to how `SPV_GOOGLE_hlsl_functionality1` subsumes `SPV_GOOGLE_decorate_string`. [[source]](https://github.com/KhronosGroup/SPIRV-Registry/blob/master/extensions/GOOGLE/SPV_GOOGLE_hlsl_functionality1.asciidoc#dependencies)

johnkslang: I think we should be saying something like "implicitly includes" (this one automatically includes the other one) rather than "subsumes" (this one is  bigger than the other, making the other one not necessary).

ehsannas: SGTM

PR #50, opened 2019-09-18
Add SPV_KHR_physical_storage_buffer.

johnkslang:


dneto0:
I have https://github.com/KhronosGroup/SPIRV-Registry/pull/51 open to modify SPV_EXT_physical_storage_buffer.  Once that is merged I can apply the same change here.  Or you can apply that change here first.  Either way is fine with me.

johnkslang:
> I have #51 open to modify SPV_EXT_physical_storage_buffer. Once that is merged I can apply the same change here. Or you can apply that change here first. Either way is fine with me.

Done.

PR #55, opened 2019-12-18
Initial release of SPV_INTEL_kernel_attributes

mkinsner:


johnkslang:
Do you also want to put an entry in the README index of extensions?

mkinsner:
> Do you also want to put an entry in the README index of extensions?

Yes, good point.  I've updated the MR to include this change.

mkinsner:
@johnkslang I believe that this is ready for merge

PR #70, opened 2020-05-27
SPV_AMD_texture_gather_bias_lod: more explicit validation rule

nhaehnle:
One could argue that it's implied by some of the language,
but let's be 100% clear that Lod and Bias cannot be used
together.


nhaehnle:
@johnkslang could you please merge this?

PR #86, opened 2020-11-23
Updates to final ray tracing extensions

dgkoch:
Final versions of SPV_KHR_ray_tracing and SPV_KHR_ray_query


dgkoch:
cc @raunraun @johnkslang

PR #117, opened 2021-06-16
Add the SPV_EXT_shader_atomic_float16_add extension

jekstrand:


johnkslang:
I think this still needs to be added to the README index.

jekstrand:
Done

extensions/EXT/SPV_EXT_shader_atomic_float16_add.asciidoc
```diff
@@ -0,0 +1,138 @@
...
+Contributors
+------------
+
+- Jason Ekstrand, Intel
+- Ben Ashbaugh, Intel
+
+Status
+------
+
+- Draft
```
bashbaug, line 25: We should fix this to have the approval dates also, since this is no longer a `Draft`.

bashbaug (commented):
We should fix the extension "Status" also, otherwise LGTM.

extensions/EXT/SPV_EXT_shader_atomic_float16_add.asciidoc
jekstrand: Yup.  Fixed.  While I was fixing it, I looked at SPV_EXT_shader_atomic_float_add to see what to do and it was still marked draft too. :joy:

Issue #124, opened 2021-08-09
Upcoming renaming of master -> main branch planned for late August 2021


oddhack:
@johnkslang @dneto0 @raunraun @HansKristian-Work and other interested parties: heads-up that we propose to rename the default branch of this repository to 'main' in late August 2021, following emerging Khronos & github practice (for Khronos members, see internal khronos-general issue 106). This will have very little impact, although if you have a checked-out repository clone, you may want to follow the simple instructions github will pop up on the repository webpage describing how to do the same locally.

If there is a concern about this, please raise it here by 2021-08-23.

johnkslang:
What will happen to all the links in the README index section that contain paths like ".../blob/master/..."?  Will they continue to work, or should they all also change to ".../blob/main/..." at the same time?

oddhack:
Github rewrites them for us! For example if you go to a repo where this has already been done, like https://github.com/KhronosGroup/SPIR-Registry/blob/master/CODE_OF_CONDUCT.md, the page will come up with 'master' rewritten to 'main'. That said it is still advisable to change all deep links from 'master' to 'main' eventually, but there's no urgency to it.

oddhack:
Default branch has been renamed. I made the edits to the previewer links in README.md in #126, which should be accepted soon.

PR #126, opened 2021-08-24
Rename extension preview links from master -> main

oddhack:
The existing links to master from README.md still work after renaming
the default branch to main (see #124). But unlike some other cases, github does not
actually rewrite the URLs, just does the branch aliasing internally. So it's a good idea to explicitly rename them sooner rather than later.


oddhack:
@raunraun you seem to be handling accepting MRs so assigning over to you. Note you can sanity check the changes in this PR by going to the 'Files changed' listing, 'View file' on README.md, and following links, which resolve to the same files as the existing links.

PR #143, opened 2022-05-03
update the HTML spec for SPV_KHR_uniform_group_instructions

bashbaug:
It looks like the HTML spec for SPV_KHR_uniform_group_instructions was built from an older asciidoc spec source and using different asciidoc style sheet.  This PR rebuilds the HTML spec with the latest asciidoc spec source in this repo and using the standard SPIR-V style sheet.


bashbaug:
In case it helps for review, here is the [old HTML spec](http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_uniform_group_instructions.html), and here is the [new HTML spec](http://htmlpreview.github.io/?https://github.com/bashbaug/SPIRV-Registry/blob/update-uniform-group-instructions-html/extensions/KHR/SPV_KHR_uniform_group_instructions.html) for comparison.

PR #148, opened 2022-06-21
An extension that relaxes the constant address space requirement of printf format strings and string args

pjaaskel:
The key motivation for this extension is to minimize the need for modifications of codes using printf when moving between host and device, and to support running CUDA/HIP-based kernels which call a more relaxed printf (that in fact delegates from the device to host printf, whatever that happens to be).

The OpenCL C part is in https://github.com/KhronosGroup/OpenCL-Docs/pull/807.


README.md
```diff
@@ -106,6 +106,7 @@ Khronos SPIR-V Registry](https://www.khronos.org/registry/spir-v/).
 1. [SPV_INTEL_split_barrier                 ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/INTEL/SPV_INTEL_split_barrier.html)
 1. [SPV_KHR_ray_cull_mask                   ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_ray_cull_mask.html)
 1. [SPV_KHR_fragment_shader_barycentric     ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_fragment_shader_barycentric.html)
+1. [SPV_POCL_relaxed_printf_string_address_space      ]( http://htmlpreview.github.io/?https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/POCL/SPV_POCL_relaxed_printf_string_address_space.html)
```
raunraun, line 109: README has different extension name from the asciidoc

pjaaskel: Thanks, fixed.

PR #153, opened 2022-07-03
Update clspv reflection non-semantic instruction set

kpet:
Add support for:

- Pointer arguments
- Program scope variables
- Image channel order and data type queries

Signed-off-by: Kévin Petit <kpet@free.fr>

### DIFF
--- a/nonsemantic/NonSemantic.ClspvReflection.asciidoc
+++ b/nonsemantic/NonSemantic.ClspvReflection.asciidoc
@@ -1,7 +1,7 @@
 SPIR-V Non-Semantic Clspv Reflection Instructions
 =================================================
 
-Version 2
+Version 3
 
 Contact
 -------
@@ -19,7 +19,7 @@ Contributors
 Notice
 ------
 
-Copyright (c) 2020 The Khronos Group Inc. Copyright terms at
+Copyright (c) 2020-2022 The Khronos Group Inc. Copyright terms at
 http://www.khronos.org/registry/speccopyright.html
 
 Status
@@ -32,8 +32,8 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2021-10-25
-| Revision           | 2
+| Last Modified Date | 2022-06-26
+| Revision           | 3
 |========================================
 
 Dependencies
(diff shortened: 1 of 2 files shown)


nonsemantic/NonSemantic.ClspvReflection.asciidoc
```diff
@@ -677,6 +677,243 @@ specialization id for the value. +
 'Size'
 |=====
 
+[cols="6"]
+|=====
+6+|[[ArgumentPointerPushConstant]]*ArgumentPointerPushConstant* +
```
alan-baker, line 682: Please add the version info to these additions, e.g.
>Missing before version 3

Separately, I realized there's a bad token in `SpecConstantSubgroupMaxSize` where there is a stray `+`.

nonsemantic/NonSemantic.ClspvReflection.asciidoc
```diff
@@ -677,6 +677,243 @@ specialization id for the value. +
...
+'PointerOffset' |
+'<id>' +
+'PointerSize'
+|=====
+
+[cols="5"]
+|=====
+5+|[[ImageArgumentInfoChannelOrderPushConstant]]*ImageArgumentInfoChannelOrderPushConstant* +
+ +
+Declares a *PushConstant* location to pass the channel order of the image that
```
alan-baker, line 797: I think maybe there should be a link for channel order and format entries, but I'm not sure to which spec (probably 5.3.1.1 of the CL spec for channel order).

alan-baker (changes requested):
Please add the version descriptions in the new entries.

nonsemantic/NonSemantic.ClspvReflection.asciidoc
kpet: Why not. I've linked the OpenCL C spec since the values passed around here will be the CLK ones rather than the API-side definitions but if you prefer linking the API spec that does have more background on image formats instead, I'm equally fine with that.

kpet:
@alan-baker  Thanks for the review. I think I've addressed all your comments. PTAL.

alan-baker (approved):

PR #164, opened 2022-08-25
A Non-Semantic Extension to Programmatically Trigger Break Point

qingyuanzNV:
This extension introduces a non-semantic extension instruction DebugBreak as a hint for any attached shader debugger to hit a breakpoint programmatically. It is meant to provide similar feature like __debugbreak() in MSVC.

### DIFF
--- a/nonsemantic/NonSemantic.DebugBreak.asciidoc
+++ b/nonsemantic/NonSemantic.DebugBreak.asciidoc
@@ -1,0 +1,93 @@
+SPIR-V NonSemantic DebugBreak Instructions
+===========================================
+
+Version 1.00
+
+Contact
+-------
+
+To report problems with this extension, please open a new issue at:
+
+https://github.com/KhronosGroup/SPIRV-Headers
+
+Contributors
+------------
+
+- Qingyuan Zheng, NVIDIA Corporation
+- Ashwin Lele, NVIDIA Corporation
+- Jeff Bolz, NVIDIA Corporation
+
+Notice
+------
+
+Copyright (c) 2019-2022 The Khronos Group Inc. Copyright terms at
+http://www.khronos.org/registry/speccopyright.html
+
+Status
+------
+
+- Not Yet Approved by the SPIR-V Working Group
+
+Version
+-------
+
+[width="40%",cols="25,25"]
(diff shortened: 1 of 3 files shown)


nonsemantic/NonSemantic.DebugBreak.asciidoc
```diff
@@ -0,0 +1,93 @@
...
+2. Binary Form
+---------------
+
+[cols="1,1,2*3",width="100%"]
+|=====
+3+|[[DebugBreak]]*DebugBreak* +
+ +
+Hint the attached shader debugger to trigger a breakpoint and
+pause the execution. If no debugger is available, this instruction
+should either be ignored or terminte the program. +
```
r-potter, line 68:
```suggestion
should either be ignored or terminate the program. +
```

nonsemantic/NonSemantic.DebugBreak.asciidoc
```diff
@@ -0,0 +1,93 @@
...
+| Revision           | 1
+|========================================
+
+Dependencies
+------------
+
+This extension is written against the SPIR-V Specification,
+Version 1.6 Revision 2.
+
+This instruction set requires SPIR-V 1.0.
```
dgkoch, line 46: Should this be 1.6 (which is the version that added non-semantic support to core)

johnkslang:
Should it require the non-semantic instruction set extension instead of 1.6?

dneto0:
> Should it require the non-semantic instruction set extension instead of 1.6?

I suspect this may be driven by how the tests are written. If they assume SPIR-V 1.6 then that's enough.
Also, it's a nice nudge into the future.

dneto0:
> > Should it require the non-semantic instruction set extension instead of 1.6?
> 
> I suspect this may be driven by how the tests are written. If they assume SPIR-V 1.6 then that's enough. Also, it's a nice nudge into the future.

Oh, I didn't see the resolved conversation.  The reliance on 1.6 may be somewhat accidental. I'll step back now.

qingyuanzNV:
I've updated the requirements to SPIR-V 1.0 and SPV_KHR_non_semantic_info.

PR #169, opened 2022-09-16
Allow at most one task payload variable per TaskEXT entry point

rg3igalia:
This appears to be the desired behavior and matches the language in the issues section, which explains that there can be only one task payload variable per entry point, without specifying TaskEXT or MeshEXT.

/cc @pmistryNV @dgkoch @HansKristian-Work


pmistryNV (approved):
Looks good to me

PR #170, opened 2022-09-18
Add SPV_QCOM_image_processing

wooyoungqcom:


extensions/QCOM/SPV_QCOM_image_processing.asciidoc
```diff
@@ -0,0 +1,285 @@
...
+Notice
+------
+
+Copyright (c) 2020 The Khronos Group Inc. Copyright terms at
+http://www.khronos.org/registry/speccopyright.html
+
+Status
+------
+
+- Approved by the SPIR-V Working Group: 2022-09-07
```
johnkslang, line 33: I think this might be copied from another extension, what's the actual status?

wooyoungqcom: Fixed it to "Final"

PR #178, opened 2022-12-08
Update clspv reflection non-semantic instruction set

callumfare:
Add support to the clspv reflection non-semantic instruction set for:
* Printf builtin
* Device address push constants for module-scope constants and program-scope global variables
* NumArguments, Flags, and Attributes for the Kernel instruction

This contribution is being made by Codeplay on behalf of Samsung.


alan-baker (changes requested):

callumfare:
@alan-baker Thanks for the review, I've addressed your comments including adding the `NumArguments` operand. I also corrected the description of the printf buffer - it doesn't need to be zero-initialized, only the first 4 bytes need to be zeroed.

callumfare:
I've addressed the review feedback. I've also added a change from @FMarno which adds another operand to support clGetKernelInfo with `CL_KERNEL_ATTRIBUTES`

alan-baker (approved):

PR #179, opened 2022-12-09
Add INTEL/SPV_INTEL_fpga_latency_control.asciidoc

danielz1intel:
Add INTEL/SPV_INTEL_fpga_latency_control.asciidoc.


bashbaug (commented):
Hi, can you please:

* Include an HTML rendering for this extension. I can help with this if needed.
* Update the README to include a link to the HTML rendering?

danielz1intel:
> Hi, can you please:
> 
> * Include an HTML rendering for this extension. I can help with this if needed.
> * Update the README to include a link to the HTML rendering?

Updated, please review, thanks!

extensions/INTEL/SPV_INTEL_fpga_latency_control.asciidoc
```diff
@@ -0,0 +1,158 @@
...
+[options="header"]
+|====
+2+^| Capability ^| Implicitly Declares
+| 6171 | FPGALatencyControlINTEL |
+|====
+--
+
+=== Validation Rules
+
+* *Note* that it is invalid for two labeled instructions to be decorated with *LatencyControlLabelINTEL* using the same _Latency Label_.
```
bashbaug, line 143: Mentioned in last week's SPIR-V call: this condition is already described generally in the main SPIR-V specification so it does not need to be duplicated here.  From [3.20 Decoration](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#_decoration):

> It is invalid to decorate any given id or structure member more than one time with the same decoration, unless explicitly allowed below for a specific decoration.

It would be best to remove this validation rule to eliminate the duplication.

bashbaug (approved):
Thanks, latest version LGTM.

PR #184, opened 2023-01-18
Mark SPV_NV_shader_subgroup_partitioned as complete

dgkoch:
Pointed out in https://github.com/KhronosGroup/SPIRV-Headers/issues/310


jeffbolznv:
LGTM

PR #185, opened 2023-02-01
Add INTEL/SPV_INTEL_fpga_buffer_location.asciidoc

danielz1intel:
Add INTEL/SPV_INTEL_fpga_buffer_location.asciidoc


bashbaug (approved):
LGTM, thanks!  Needs a non-Intel reviewer to merge.

extensions/INTEL/SPV_INTEL_fpga_buffer_location.asciidoc
```diff
@@ -0,0 +1,122 @@
...
+== Dependencies
+
+This extension is written against the SPIR-V Specification,
+Version 1.6 Revision 2.
+
+This extension requires SPIR-V 1.0.
+
+== Overview
+
+This extension adds a function parameter decoration that is useful for FPGA targets.  This decoration indicates that a particular global memory pointer can only access a particular physical memory location.  Knowing this information at compile time can allow FPGA compilers to generate load store units of lower area for accesses done through such a pointer.
```
MrSidims, line 43: Taking an account https://github.com/KhronosGroup/SPIRV-LLVM-Translator/pull/1801/files - the overview should be updated

extensions/INTEL/SPV_INTEL_fpga_buffer_location.asciidoc
```diff
@@ -0,0 +1,122 @@
...
+=== Decoration
+
+Modify Section 3.20, Decoration, adding these rows to the Decoration table:
+
+--
+[options="header"]
+|====
+2+^| Decoration | Extra Operands | Enabling Capabilities
+| 5921 | *BufferLocationINTEL* + 
+Apply only to a pointer. Indicates that the pointer must only point into the physical memory identified by the subsequent literal number operand. The exact semantics of the literal number are implementation defined. | Literal +
```
MrSidims, line 82: Shouldn't it be:

```suggestion
Apply only to a pointer. Indicates that the pointer and all of its aliases must only point into the physical memory identified by the subsequent literal number operand. The exact semantics of the literal number are implementation defined. | Literal +
```
?

extensions/INTEL/SPV_INTEL_fpga_buffer_location.asciidoc
broxigarchen: Right. I think this should be both function parameter decoration and Instruction parameter decoration.

There are some history here. The function parameter decoration is already checked in into the code but the spec is not yet posted at that time... So this spec actually covers both

extensions/INTEL/SPV_INTEL_fpga_buffer_location.asciidoc
broxigarchen: I think it makes sense but I am not sure if this is necessary. Wouldn't alias the word itself already implicitly indicates that these two pointers are equivalent?

extensions/INTEL/SPV_INTEL_fpga_buffer_location.asciidoc
broxigarchen, line 43:
```suggestion
This extension adds a decoration for a function parameter or a pointer argument of memory access instructions that is useful for FPGA targets.  This decoration indicates that a particular global memory pointer can only access a particular physical memory location.  Knowing this information at compile time can allow FPGA compilers to generate load store units of lower area for accesses done through such a pointer.  
```

broxigarchen: maybe something like this?

extensions/INTEL/SPV_INTEL_fpga_buffer_location.asciidoc
MrSidims: After some thinking - I no longer have objections of leaving this as is.

extensions/INTEL/SPV_INTEL_fpga_buffer_location.asciidoc
danielz1intel: Please review the recent commit to see if the displayed is appropriate. Thanks!

MrSidims: Moved the discussion internally. Once internal discussion resolved please don't forget to update generated .html file here.

danielz1intel: Adjusted the docs in this PR after internal adjustments.

MrSidims (approved):
LGTM

broxigarchen (approved):

PR #188, opened 2023-02-28
Update Execution Mode argument name to match name in the SYCL extension

tiwaria1:
The argument name for the mode `RegisterMapInterfaceINTEL` in the SPIRV spec does not match the name in the [SYCL extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/proposed/sycl_ext_intel_fpga_kernel_interface_properties.asciidoc) 
 (see property `register_map_interface`). This PR updates the SPIRV extension to align the names.

### DIFF
--- a/extensions/INTEL/SPV_INTEL_kernel_attributes.asciidoc
+++ b/extensions/INTEL/SPV_INTEL_kernel_attributes.asciidoc
@@ -154,10 +154,10 @@ _StallFreeReturn_
 
 | 6160 | *RegisterMapInterfaceINTEL* +
 Indicates that the kernel has a single register based interface that is shared across all kernel control signals and kernel arguments.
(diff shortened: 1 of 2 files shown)


tiwaria1:
@MrSidims  Could you review this minor change (I'll update the translator PR too)

bashbaug:
If we do this I think we'll want to update the grammar file and headers also (https://github.com/KhronosGroup/SPIRV-Headers/pull/317).

bashbaug (changes requested):
If we do this please build an HTML spec with these changes as well.

tiwaria1:
@bashbaug I updated the html spec and the grammar file on the headers PR. Thank you for reviewing so quickly.

bashbaug (approved):
Thanks, LGTM!

Issue #181, opened 2022-12-16
Update README to document HTML generation best practices?


bashbaug:
I think it would be a good idea to describe best practices when generating the HTML artifacts for an asciidoc extension.

I've been using something like the following, which matches the HTML output for many (but not all) of the HTML extensions in this repo:

```sh
$ asciidoc -b html5 -a icons -a toc2 -a toclevels=1 -o SPV_my_extension.html SPV_my_extension.asciidoc
```

There is probably an equivalent that uses the more modern asciidoctor toolchain but I haven't had the need to figure out what it is... yet.

Would this be helpful?  Should we document this in the README or using some other mechanism?  Should we include a Makefile in this repo?

dgkoch:
IMO we should include a Makefile in this repo (and probably document it as well, but definitely a Makefile so that the html files can be (re-)generated in place).

bashbaug:
Fixed by #189.

PR #189, opened 2023-03-04
add a Makefile to compile asciidoc files to HTML

bashbaug:
Possible fix for #181 

This PR adds a Makefile to create an HTML spec from asciidoc spec source in the extensions directory.

I haven't updated the README with these instructions but perhaps the Makefile is good enough?  Or should we include best practices in a separate document?  The README is primarily an index with links to HTML specs right now.

Note also this is still using the older asciidoc toolchain but we can shift to asciidoctor if we'd like, we just need to figure out which command line to use.


dgkoch:
This looks pretty similar to my local script so +1 for that. 

I'd suggest we should document it either a BUILD.md or CONTRIBUTING.md and include a link to that from the top of the README.md

bashbaug:
Thanks, I added a BUILD.md and linked it from the README.

dgkoch (approved):
LGTM

PR #193, opened 2023-03-16
Add INTEL/SPV_INTEL_fp_max_error.asciidoc

danielz1intel:
Add INTEL/SPV_INTEL_fp_max_error.asciidoc


bashbaug (commented):
LGTM.

Just observing: The HTML was built using Asciidoctor rather than Asciidoc and the steps outlined in BUILD.md and the Makefile, but it looks nice.

Needs a non-Intel reviewer before merging.

extensions/INTEL/SPV_INTEL_fp_max_error.asciidoc
```diff
@@ -0,0 +1,135 @@
...
+* Shuo Niu, Intel
+* Daniel Zhang, Intel
+
+== Notice
+
+Copyright (c) 2023 Intel Corporation.  All rights reserved.
+
+== Status
+
+Final draft
```
bashbaug, line 25: Minor, mentioning since it came up in Wednesday's teleconference: are we OK with the status being "Final draft" or would we rather this be something like "Shipping"?

```suggestion
* Shipping
```

PR #194, opened 2023-03-20
Fix DebugTypeBasic typo in NonSemantic.Shader.DebugInfo.100

MrSidims:
DebugBasicType -> DebugTypeBasic


baldurk:
Looks like a simple typo indeed, seems fine to me.

PR #206, opened 2023-06-21
Publish SPV_EXT_image_raw10_raw12

kpet:
Used by the following OpenCL extension: https://github.com/KhronosGroup/OpenCL-Docs/pull/919

### DIFF
--- a/extensions/EXT/SPV_EXT_image_raw10_raw12.asciidoc
+++ b/extensions/EXT/SPV_EXT_image_raw10_raw12.asciidoc
@@ -1,0 +1,98 @@
+SPV_EXT_image_raw10_raw12
+=========================
+
+Name Strings
+------------
+
+SPV_EXT_image_raw10_raw12
+
+Contact
+-------
+
+To report problems with this extension, please open a new issue at:
+
+https://github.com/KhronosGroup/SPIRV-Headers
+
+Contributors
+------------
+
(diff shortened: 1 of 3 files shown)


extensions/EXT/SPV_EXT_image_raw10_raw12.asciidoc
```diff
@@ -0,0 +1,98 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Modify Section 3.13, "Image Channel Data Type", adding these rows to the table:
+
+--
+[options="header"]
+|====
+2+^| Image Channel Data Type ^| Enabling Capabilities
+| 19 | *UnsignedIntRaw10EXT* | *Kernel*
+| 20 | *UnsignedIntRaw12EXT* | *Kernel*
```
bashbaug, line 80: I was going to suggest that these new "image channel data types" shouldn't be guarded by the Kernel capability, but it looks like all of the other image channel data types are also guarded by the Kernel capability, as is the related [OpImageQueryFormat](https://registry.khronos.org/SPIR-V/specs/unified1/SPIRV.html#OpImageQueryFormat) instruction, so this is _probably_ OK as-is.

Adding a review comment for the record, and just in case somebody else has a similar reaction, but no action required.

bashbaug (approved):

PR #207, opened 2023-06-21
Publish SPV_KHR_cooperative_matrix

kpet:


extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
```diff
@@ -0,0 +1,602 @@
...
+------
+
+Copyright (c) 2020-2023 The Khronos Group Inc. Copyright terms at
+http://www.khronos.org/registry/speccopyright.html
+
+Status
+------
+
+- Approved by the SPIR-V Working Group: 2023-05-03
+- Not Yet Approved by the Khronos Board of Promoters: (To become approved, with an approval date, e.g. 2018-01-30)
```
alan-baker, line 51: This should be updated with the ratification date.

kpet: Good catch! I'd made the change locally (hence why it's in the HTML) but did not commit it.

PR #214, opened 2023-07-28
Add SPV_AMDX_shader_enqueue

Tobski:


Tobski:
@johnkslang can you merge this ASAP as discussed, so that we have links working etc.?

johnkslang:
Possibly needs hooking into the README, and maybe something else instead of "may:", which is not a SPIR-V macro, and is perhaps Vulkan specific.

PR #217, opened 2023-08-30
Add spec for SPV_NV_displacement_micromap.

alelenv:


extensions/NV/SPV_NV_displacement_micromap.asciidoc
```diff
@@ -0,0 +1,297 @@
...
+
+- Eric Werness, NVIDIA
+- Ashwin Lele, NVIDIA
+- Pyarelal Knowles, NVIDIA
+- Christoph Kubisch, NVIDIA
+
+Status
+------
+
+- Provisional
```
bashbaug, line 26: I don't have a super strong opinion about this, but we should think about any conventions and policies for "provisional" extensions.  For example, the `SPV_AMDX_shader_enqueue` extension includes `X` in its name and has a big disclaimer describing what is intended by "provisional".  Should we include similar text here?

extensions/NV/SPV_NV_displacement_micromap.asciidoc
```diff
@@ -0,0 +1,297 @@
...
+|========================================
+| Last Modified Date | 2023-08-01
+| Revision           | 1
+|========================================
+
+Dependencies
+------------
+
+This extension is written against the SPIR-V Specification,
+Version 1.1 Revision 4.
```
bashbaug, line 41: This doesn't _need_ to be updated, but the SPIR-V 1.1 spec is pretty old.  Would it make sense to baseline this extension against the SPIR-V 1.6 spec instead?

bashbaug (commented):
Can you please link this extension from the README as well?

pknowlesnv:
> Missing addition to table of contents?

Sure. Can you point me to an existing example of how this is done in asciidoc (e.g. another extension that adds what you have in mind)?

extensions/NV/SPV_NV_displacement_micromap.asciidoc
pknowlesnv: Sounds good, thanks! I saw the same blurb e.g. in the provisional [SPV_KHR_ray_query](https://raw.githubusercontent.com/KhronosGroup/SPIRV-Registry/798acbe508c32db0f4619cc1844e8762bab945a6/extensions/KHR/SPV_KHR_ray_query.asciidoc) (older version of the current file). We'd like to do the same and avoid an intermediate X version of the extension - I think that's what you're suggesting anyway.

extensions/NV/SPV_NV_displacement_micromap.asciidoc
pknowlesnv: Good point, thanks! This should require 1.4 anyway, to match
[SPV_KHR_ray_tracing](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_ray_tracing.asciidoc). Written-against might not matter as much but would be good to at least match ray tracing.

bashbaug:
> Sure. Can you point me to an existing example of how this is done in asciidoc (e.g. another extension that adds what you have in mind)?

I think John is making the same comment that I did - that this PR should also update the README.  Basically, you'll add this extension to the end of the list.  You can have a look at any recently published extension for an example, such as https://github.com/KhronosGroup/SPIRV-Registry/pull/207 - pay attention to the changes to README.md.  Let us know if this is still unclear - thanks!

alelenv:
Thanks, I have updated, let me know if any further feedback

extensions/NV/SPV_NV_displacement_micromap.asciidoc
```diff
@@ -0,0 +1,297 @@
...
+| *RayTracingDisplacementMicromapNV*
+|====
+--
+
+(Modify Section 3.31, Capability, adding a row to the Capability table) ::
++
+--
+[cols="3^.^,6,25,15^,15",options="header",width = "100%"]
+|====
+2+^.^| Capability|Enabled by Extension
```
alan-baker, line 169: This table has columns `Capability` and `Enabled by Extension`, but usually those columns are `Capability` and `Implicitly Declares`. The headers for this extensions seem to imply that these capabilities would declare the `Shader` capability (which is unsurprising). I'd suggest modifying this table to match that. These capabilities are introduced by this extension so it doesn't add information to say they are enabled by it too.

alan-baker (changes requested):

alan-baker (approved):
Looks ok to me now

bashbaug:
Merging as discussed in the September 20th teleconference.

PR #221, opened 2023-10-24
Clarifications to SPV_KHR_cooperative_matrix

### LINKED ISSUE #208: SPV_KHR_cooperative_matrix initial specification clarifications
We recently merged the first public revision of the SPV_KHR_cooperative_matrix specification. I'm creating this issue to record/bundle clarification requests and proposals for an upcoming revision of the specification.

1. The language we use to describe the _Memory Operand_ operand to **OpCooperativeMatrixLoadKHR** and **OpCooperativeMatrixStoreKHR** does not make it clear that multiple words can be used. We should align the language to that used for **OpLoad**, **OpStore**, etc.
2. What are the type constraints for the _Pointer_ operand to **OpCooperativeMatrixLoadKHR** and **OpCooperativeMatrixStoreKHR**? Specifically, does the pointee type need to be a scalar or vector *of the same type as the matrix's component type* or can it be any scalar or vector type.
PROPOSED RESOLUTION: Constrain the pointee type to be a scalar or vector of the same type as the matrix's component type.
RESOLUTION: Type mismatches were intended and are helpful to efficiently load into shared memory. The Vulkan spec provides alignment guarantees. No change.

### LINKED ISSUE #211: SPV_KHR_cooperative_matrix and optional operands
`OpCooperativeMatrixLoadKHR` and `OpCooperativeMatrixStoreKHR` allow for 2 optional operands: stride and 'memory operand'.

When reading these two from a spir-v binary, how am I supposed to tell which of the two possible optional operands is provided? Should I assume that both are (always) either omitted or present at the same time?

kpet:
- Change the description of 'Memory Operand' operands to load/store instructions to make it clear that multiple words can be used (in line with the language used for OpLoad/OpStore).
- Require all instructions that have a cooperative matrix object as an operand or produce a cooperative matrix object result only be executed in uniform control flow within the scope of the matrix type.
- Require the constituent provided to OpCompositeConstruct to be the same for all invocations in the scope of the cooperative matrix.
- Require the operands to arithmetic and conversion instructions operating on cooperative matrix objects to be the same for all invocations within the scope of the cooperative matrix.
- Clarify when _Stride_ and _Memory Operand_ operands can be provided.
- Clarify when referring to cooperative matrix types or objects/values of that type in the description of conversion instructions.

Fixes #208
Fixes #211

### DIFF
--- a/extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
+++ b/extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
@@ -55,8 +55,8 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2023-07-26
-| Revision           | 4
+| Last Modified Date | 2023-12-06
+| Revision           | 5
 |========================================
 
 Dependencies
@@ -268,6 +268,10 @@ where all invocations in 'Scope' cooperate to compute and store the matrix. +
 'Use' must be a 'constant instruction' scalar 32-bit 'integer type' whose
 value corresponds to a 'Cooperative Matrix Use'. +
  +
+All dynamic instances of an instruction with an operand or result that is an
+object of this type must be executed such that all the invocations in the
+'Scope' instance are active or none of them are.
+ +
 1+|Capability: +
 *CooperativeMatrixKHR*
 1+| 7 | 4456 | 'Result <id>' | '<id>' +
@@ -309,14 +313,12 @@ a description of the layouts and detailed layout-specific rules. +
 'Stride' further qualifies how matrix elements are laid out in memory. It must be a
 scalar 'integer type' and its exact semantics depend on 'MemoryLayout'. +
  +
-'Memory Operand' must be a +Memory Operand+ literal. If not present, it is the
-same as specifying *None*. +
+'Memory Operand', if present, must begin with a +Memory Operand+ literal. If not
+present, it is the same as specifying the +Memory Operand+ *None*. +
  +
-For a given dynamic instance of this instruction, all operands of this
-instruction must be the same for all invocations in a given scope instance
-(where the scope is the scope the cooperative matrix type was created with).
-All invocations in a given scope instance must be active or all must be
-inactive.
+All the operands to this instruction must be dynamically uniform within every
+instance of the 'Scope' of the cooperative matrix.
+ +
 1+|Capability: +
 *CooperativeMatrixKHR*
 1+| 5+variable | 4457 | '<id>' +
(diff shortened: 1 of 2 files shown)


kpet:
Here's a first attempt at addressing the outstanding clarification requests on SPV_KHR_cooperative_matrix (cc @alan-baker, @jeffbolznv).

extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
```diff
@@ -268,6 +268,10 @@ where all invocations in 'Scope' cooperate to compute and store the matrix. +
 'Use' must be a 'constant instruction' scalar 32-bit 'integer type' whose
 value corresponds to a 'Cooperative Matrix Use'. +
  +
+All dynamic instances of an instruction with an operand or result that is an
+object of this type must be executed such that all the invocations in the
+'Scope' instance are active or none of them are.
```
jeffbolznv, line 273: I thought the consensus was to only require this for the matmul instruction?

extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
```diff
@@ -386,20 +388,26 @@ invocation when treated as a composite. +
...
 *OpConvertSToF*, *OpConvertUToF*, *OpUConvert*, *OpSConvert*, *OpFConvert*.
-Allow the use of *OpBitcast* on cooperative matrix types whose 'Component Type'
-are integer types with the same 'Width'.
+Allow the use of *OpBitcast* on objects of cooperative matrix type whose
+'Component Type' are integer types with the same 'Width'.
 The result type and value type must have the same 'Scope', number of 'Rows',
 number of 'Columns', and 'Use'.
 
+For a given dynamic instance of these instructions, all their operands must be
+the same for all invocations in a given scope instance (where the scope is the
```
jeffbolznv, line 400: I know some of this is preexisting, but I'm not sure how to interpret "the same" in this restriction. One could interpret it to mean they must be uniform.

kpet: It's a carry over from the NV extension (in case you want to change the language there too). Agree, we can probably improve on this text. How about:

>All the operands to these instructions must be dynamically uniform within the 'Scope' of the cooperative matrix.

extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
kpet: It was also the case for the load/store instructions. If we make this proposed change, it would, strictly speaking, lead to new restrictions (that arguably might have been assumed by some). For example, some of the conversion instructions will result in cross-lane data exchanges for some vendors and uniform control flow is helpful as a restriction. Arguably this restriction might not be required on all instructions but I'm not sure we lose much with it (notably, AFAICT, element-wise access still works according to the letter of this rule). Maybe best discussed in the next teleconference (or F2F)?

kpet: Discussed at the 2023/10/25 F2F: the proposed change is fine but we should capture the fact that element access needn't be performed in uniform control flow in an issue. The reason for that, i.e. that the uniform control flow requirement as proposed applies to instructions with operands that have a cooperative matrix type but not a pointer to cooperative matrix type, is subtle and easily missed.

kpet:
@alan-baker @jeffbolznv I've updated the PR to reflect the discussions at the F2F. PTAL.

kpet:
@alan-baker I'm guessing you're referring to these rules:

>All the operands to this instruction must be dynamically uniform within the 'Scope' of the cooperative matrix.

I think the intent is clear from this wording but we may be able to improve the language. How about the following:

>All the operands to this instruction must be dynamically uniform within every instance of the 'Scope' of the cooperative matrix.

kpet:
@alan-baker I've made the change. This should now be ready to go.

PR #223, opened 2023-11-15
Upstream SPV_INTEL_masked_gather_scatter

MrSidims:
This extension allows TypeVector to have a physical pointer type Component Type and introduces gather/scatter instructions. It will be useful for explicitly vectorized kernels.

SPIR-V validator adjustments will be done later.

SPIR-V Headers: https://github.com/KhronosGroup/SPIRV-Headers/pull/391

### DIFF
--- a/extensions/INTEL/SPV_INTEL_masked_gather_scatter.asciidoc
+++ b/extensions/INTEL/SPV_INTEL_masked_gather_scatter.asciidoc
@@ -37,7 +37,7 @@ Copyright (c) 2023 Intel Corporation.  All rights reserved.
 [width="40%",cols="25,25"]
 |========================================
 | Last Modified Date | 2023-09-05
-| Revision           | 2
+| Revision           | 1
 |========================================
 
 == Dependencies
@@ -286,6 +286,5 @@ Revision History
 [options="header"]
 |========================================
 |Rev|Date|Author|Changes
-|1|2022-08-19|Dmitry Sidorov|Initial revision
-|2|2023-09-05|Dmitry Sidorov|Prepare to ship
+|1|2023-09-05|Dmitry Sidorov|Prepare to ship
 |========================================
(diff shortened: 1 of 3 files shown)


extensions/INTEL/SPV_INTEL_masked_gather_scatter.asciidoc
```diff
@@ -0,0 +1,291 @@
...
+
+Revision History
+----------------
+[cols="5,15,15,70"]
+[grid="rows"]
+[options="header"]
+|========================================
+|Rev|Date|Author|Changes
+|1|2022-08-19|Dmitry Sidorov|Initial revision
+|2|2023-09-05|Dmitry Sidorov|Prepare to ship
```
bashbaug, line 290: The first published version is usually version 1, but I'm fine publishing version 2 if you are.

bashbaug (approved):
LGTM, just one minor comment regarding the extension version.

extensions/INTEL/SPV_INTEL_masked_gather_scatter.asciidoc
```diff
@@ -0,0 +1,290 @@
...
+|=====
+
+[cols="1,1,4*3",width="100%"]
+|=====
+5+|[[OpMaskedScatterINTEL]]*OpMaskedScatterINTEL* +
+ +
+Writes values from a vector to the corresponding memory address of the given vector of pointers. Memory access
+is specified by a mask instruction parameter. +
+ +
+'InputVector' is a _vector_ of values to scatter. +
```
alan-baker, line 216: Not a blocker, but OpStore puts the pointer operand first. Not sure if you want to follow that precedent.

MrSidims: Good comment. I agree, that the order should follow OpStore. The problem though is that we have implemented the instruction across the software stack and changing order now would be backward incompatible. I see two possible ways to deal with it:
1. Change Op code and the order now, leaving for us to support both old and new Op codes and orders;
2. If somebody express an interest to make this extension EXT/KHR or add the instructions to the core spec - we can change the Op code and the order then.

I'd slightly prefer to stick with the option 2, wdyt?

alan-baker: If you've implemented it already I agree it is not worth changing now. We can think about option 2 if there is ever wider adoption.

PR #227, opened 2023-12-06
Remove references to non-existent SPIR-V definitions

dgkoch:
We only define "invocation reorder" instructions in Vulkan, not SPIR-V


alan-baker (approved):

bashbaug:
Merging as discussed in the December 13th teleconference.

PR #230, opened 2024-01-22
Updates to SPV_GOOGLE_user_type

sudonatalie:


sudonatalie:
@chaoticbob Please take a look

extensions/GOOGLE/SPV_GOOGLE_user_type.asciidoc
```diff
@@ -18,6 +18,7 @@ Contributors
 - Hai Nguyen, Google
 - David Neto, Google
 - Ehsan Nasiri, Google
+- Natalie Chouinard, Google
 
 Status
 ------
```
dgkoch, line 24: Version table below this needs to be updated.

sudonatalie: Do you mean the Revision History table at the bottom of this file? I added a row to it.

dgkoch: No, I mean lines 32 and 33, which are collapsed in the diff and I can't comment on them. 
They should be updated to match the latest entry in the Revision History table.

sudonatalie: Oops, don't know how I missed that on the second look too. Thanks!

extensions/GOOGLE/SPV_GOOGLE_user_type.asciidoc
```diff
@@ -28,8 +29,8 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2019-04-22
-| Initial            | 1
+| Last Modified Date | 2024-01-25
+| Revision           | 2
```
dgkoch, line 33:
```suggestion
| Revision           | 5
```

extensions/GOOGLE/SPV_GOOGLE_user_type.asciidoc
```diff
@@ -45,13 +46,13 @@ extension.
 
 Overview
 --------
-This extension provides one new decoration, UserTypeGOOGLE, that allows HLL
+This extension provides one new decoration, UserTypeGOOGLE, that allows HLSL
```
dgkoch, line 49: I suspect this is actually supposed to be HLL for "High level languages"  (This isn't only useful for HLSL is it?)

sudonatalie: Makes sense. I think it's only used for HLSL right now (the resource types below are HLSL specific), but it wouldn't have to be limited to that.

dgkoch (approved):
LGTM now.

dgkoch:
> Thanks @dgkoch, and sorry for the excess back and forth on those changes! Can you merge this PR for me?

I would prefer to leave that to @bashbaug

s-perron (approved):

PR #231, opened 2024-01-22
Update DebugGlobalVariable's Variable field

LU-JOHN:
Update description of DebugGlobalVariable's Variable field to allow Variable to hold a DebugExpression if the associated variable was optimized out.  The DebugExpression will hold the constant value of the variable that was optimized out.


MrSidims (approved):
Ideally we should also update SPIR-V validator, but guess we need to get approvals here first.

arcady-lunarg:
Ultimately this debug info is consumed by Renderdoc so I think it might be good to hear @baldurk's opinion here.

bashbaug:
Merging as discussed in the January 24th teleconference.

PR #233, opened 2024-01-25
Add SPV_KHR_quad_control

Tobski:


bashbaug (commented):
I think this also needs the HTML spec and updates to the README.

Tobski:
@bashbaug sorry this took so long - added the html and readme entry

extensions/KHR/SPV_KHR_quad_control.asciidoc
```diff
@@ -0,0 +1,183 @@
...
+Notice
+------
+
+Copyright (c) 2024 The Khronos Group Inc. Copyright terms at
+http://www.khronos.org/registry/speccopyright.html
+
+Status
+------
+
+- Draft
```
bashbaug, line 33: So sorry I didn't notice this previously.  I think the "status" needs to be updated.  The minimum change would be:

```suggestion
- Complete
```

Or, since this is a KHR extension, we could say:

```
- Approved by the SPIR-V Working Group: 2023-12-06
- Approved by the Khronos Board of Promoters: 2024-01-19
```

Tobski: Done and rebuilt.

bashbaug (approved):
Thanks for the fast turnaround.  LGTM!

bashbaug:
Merging as discussed in the March 13th teleconference.

PR #236, opened 2024-02-07
Update HTML for all asciidoc files

cassiebeckley:
It seems that sometimes people edit the asciidoc file without rebuilding the corresponding HTML file. I've rebuilt all files; some are only minor changes because of a new asciidoc utility version.

Perhaps a GitHub workflow should be added, either to check that HTML has been rebuilt for PRs, or a semi-regularly scheduled one to send in a PR rebuilding all of these.

### DIFF
--- a/extensions/AMD/SPV_AMDX_shader_enqueue.html
+++ b/extensions/AMD/SPV_AMDX_shader_enqueue.html
@@ -2,7 +2,7 @@
 <html lang="en">
 <head>
 <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
-<meta name="generator" content="AsciiDoc 9.0.0rc1">
+<meta name="generator" content="AsciiDoc 10.2.0">
 <title>SPV_AMDX_shader_enqueue</title>
 <style type="text/css">
 /* Shared CSS for AsciiDoc xhtml11 and html5 backends */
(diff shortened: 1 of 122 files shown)


s-perron:
I'm not the right person to review this. I do not know the SPIR-V Registry's processes.

bashbaug (changes requested):
Hello!  Thank you for taking the time to create this PR.  We discussed it in our weekly working group teleconference and have a few observations:

1. While we understand it is easiest to rebuild all of the HTML files it also makes this PR difficult to review and introduces new HTML versions in many cases where the only thing that "changed" is the asciidoc version and the last modified date.  We would prefer to only update the extensions that materially changed.

2.  A few of the extensions seemed to go "backwards" in this PR, indicating that the asciidoc source is out-of-date, not the HTML specifications.

Because of these issues we are unlikely to merge the PR in this form.  Could we perhaps start with a list of extensions that appear to be out-of-sync and then work our a plan to sync them?

Your idea to introduce automation to ensure a consistent look-and-feel across the HTML specifications is interesting and seems like a good long-term direction, if this is something you might be able to help with.

bashbaug:
Hi @cassiebeckley, we've been working on a workflow to automatically build and publish the HTML specifications on every commit, see https://github.com/KhronosGroup/SPIRV-Registry/issues/270.  It's not quite working yet, in the sense that our README still links to the HTML specifications in this repo, but we're getting close.  Will this work for you?  Thanks!

bashbaug:
Hi @cassiebeckley, we're now automatically building and publishing HTML files from the asciidoctor source in this repo for every commit.  Since the HTML files are now guaranteed to be in sync with the asciidoctor source I'm going to close this PR.  Thanks for the suggestion and let us know if you find any other discrepancies!

PR #239, opened 2024-02-14
Clarify FPFastMath bit meanings

### LINKED ISSUE #237: Description of float_controls2 bits Contract, Reassoc, Transform unclear
From an internal report:
> Would it be possible to adjust the wording here at some point?  We have had some difficulty where it had not been properly understood that contracting, reassociating and transforming were conjunctive operations, i.e. an instruction is contracted with its operands, or reassociated with its operands, or transformed with its operands.  Instead, it was thought that the instruction could be contracted, and its operands could be contracted too (somewhat independently). Perhaps it may read better with "with":
>
>>Allows a floating-point operation to be contracted with any operation(s) producing its operands.
>
>>Allows a floating-point operation to be reordered with respect to any operation(s) producing its operands, according to real-number associativity rules.
>
>>Allows a floating-point operation to be transformed with respect to any operations(s) producing its operands, according to real-number rules.

gnl21:
Contract, Reassoc and Transform bits control how an operation can be treated in combination with its arguments. It was not intended that decorating an operation with 'Contract' allows its operands to be contracted independently with other instructions.

Fixes #237.

### DIFF
--- a/extensions/KHR/SPV_KHR_float_controls2.asciidoc
+++ b/extensions/KHR/SPV_KHR_float_controls2.asciidoc
@@ -163,20 +163,20 @@ Add the following rows to the FP Fast Math Mode table:
 |====
 2+^.^| FP Fast Math Mode| <<Capability,Enabling Capabilities>>
 | 0x10000 | *AllowContract* +
-Allows a floating-point operation and any operation(s) producing its operands
-to be contracted. Rounding steps may be eliminated or may preserve higher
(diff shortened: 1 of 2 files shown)


bashbaug (commented):
LGTM - I think you will want to regenerate the HTML spec also, though.

alan-baker (approved):

bashbaug:
Merging as discussed in the February 14th teleconference.

PR #243, opened 2024-03-08
Fix formatting for SPV_KHR_float_controls2

gnl21:
One of the tables was missing a header, so the whole contents of the table was being treated as a header.


bashbaug (approved):
Thanks for fixing this, LGTM.

There is one more minor issue regarding the icon (?) used for "notes", which appears to be broken:

image

Is this only a problem on my end?

gnl21:
No, I see the broken image link for the notes as well, but I wonder if it's just the github preview of the HTML that's causing it? I'm not sure where (if?) the HTML versions are officially available.

bashbaug:
Merging as discussed in the March 13th teleconference.

Issue #245, opened 2024-03-15
Options for fixing Note icons


bashbaug:
From a recent review comment: https://github.com/KhronosGroup/SPIRV-Registry/pull/243#pullrequestreview-1932696242

The "Note" icon in at least one of our specs is a broken image link.  I think I know what's happening to cause this.  By default, if you define an "icons" attribute, asciidoctor looks for images in a specified directory ([link](https://docs.asciidoctor.org/asciidoc/latest/macros/icons/)).  If the images don't exist here, the icons show up as broken links, which explains the behavior we are seeing.

Here are some options to fix this:

1. We add the images in this repo somewhere, and point the icons directory to it.
2. We don't define the "icons" attribute.  This will omit the image icon, and the "Note" will show up as text only.  Needs a Makefile update in this repo.
3. We define the "icons" attribute to be "font".  This seems to work for me and produces nice output, but only for "asciidoctor", and not for "asciidoc".  Our "official" toolchain still uses "asciidoc" though we really ought to update to "asciidoctor" at some point in the future.

(2) is the path of least resistance.

In the meantime, @gnl21, if you want to "fix" SPV_KHR_float_controls2, here are a few command lines that I think will work:

```
asciidoc -b html5 -a toc2 -a toclevels=1 extensions\KHR\SPV_KHR_float_controls2.asciidoc
asciidoctor -b html5 -a toc2 -a toclevels=1 extensions\KHR\SPV_KHR_float_controls2.asciidoc 
asciidoctor -b html5 -a icons=font -a toc2 -a toclevels=1 extensions\KHR\SPV_KHR_float_controls2.asciidoc
```

Naghasan:
FYI, [Vulkan](https://github.com/KhronosGroup/Vulkan-Docs/blob/main/vkspec.adoc?plain=1#L30) and SYCL uses `icons=font` by default, but it is embedded in the main docs. We should probably define a settings document extension would import to unify the rendering of extensions.

> Our "official" toolchain still uses "asciidoc" though we really ought to update to "asciidoctor" at some point in the future.

The main spec is already using asciidoctor and the khronos docker image only provides this tool. We probably need to update the build doc here.

bashbaug:
I think we can close this now.  As long as you use the updated toolchain added in #247 the "note" icon should appear correctly.

PR #247, opened 2024-03-28
Update Makefile to better align with main spec generator

Naghasan:
- Recommend use of asciidoctor rather than asciidoc
- Use similar invocation as main SPIR-V specs
- Add shorthand to just name the extension rather than the whole path
- Add rule to run steps inside the asciidoctor-spec docker image
- Update the BUILD.md

Should address comments in https://github.com/KhronosGroup/SPIRV-Registry/issues/245

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -4,7 +4,7 @@ This document describes how to create HTML specifications for this repo.
 
 ## Asciidoc Specification Source
 
-Khronos recommends using [Asciidoc](https://asciidoc-py.github.io/index.html) to author SPIR-V extensions.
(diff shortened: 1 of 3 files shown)


bashbaug:
Mechanically this all looks good to me, thanks!

The SPIR-V spec is built with a slightly different style sheet, though.  This is most apparent in the font that gets used (sans-serif vs. serif).  Is this something we should try to align, while we're making these changes?

Naghasan:
> The SPIR-V spec is built with a slightly different style sheet, though.

True, I'll add that as well.

PR #257, opened 2024-06-12
alphabetize extensions in the README

### LINKED ISSUE #242: List Extensions in Alphabetical Order in the README
We should consider listing extensions in alphabetical order in the README instead of the current "numerical" order (based on date).

* The most common issue we encounter when merging PRs to this repo is a merge conflict in the README when two extensions each add themselves as the last item in the list.  Switching to alphabetical order would not completely eliminate this issue in theory, but it should make it much less likely to occur in practice.
* If I am searching for an extension by name it is obviously much easier to find if the list is in alphabetical order.  Currently, I almost always need to use my browser's search feature to find the extension I am looking for.
* Listing extensions in alphabetical order also makes it more likely that related extensions will be listed together, which may help discoverability.
* Some Khronos APIs assign numerical IDs to extensions (I believe OpenGL still does, maybe others), but SPIR-V does not, so the number of the extension in the list does not have any specific meaning.
* We can always use git history if needed, to reconstruct the order extensions were added to the README and this repo.

bashbaug:
fixes #242 

We merged a PR recently which inserted an extension into the middle of the list in the README, so we may as well switch from the old numbered list to an alphabetized list.  This should avoid most common merge conflicts when multiple extensions are added around the same time and make it eaiser to find an extension by name in the list.

Possible additional improvements we could consider:

1. Do we want to separate out KHR extensions or keep them in the same list?
2. Do we want to separate out EXT extensions?


Naghasan (approved):
LGTM, I'd be in favour of separating KHR and EXT, but it is minor IMO

bashbaug:
I've separated out the KHR and EXT extensions, as discussed.

bashbaug:
Merging as discussed in the July 17th teleconference.

PR #264, opened 2024-08-07
Clarify sign/zero-extension behavior in coopmatmuladd

jeffbolznv:


kpet (approved):

Naghasan (approved):

extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
```diff
@@ -415,7 +415,10 @@ be dynamically uniform within the 'Scope' of the cooperative matrix type.
 Linear-algebraic matrix multiply of 'A' by 'B' and then component-wise
 add 'C'. The order of the operations is implementation-dependent. The
 internal precision of floating-point operations is defined by the client
-API. Integer operations used in the multiplication of 'A' by 'B' are
+API. If any of the *Matrix{A,B,C}SignedComponents* operands are present,
```
wooyoungqcom, line 418: Is the suffix "KHR" missing ?

extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
```diff
@@ -415,7 +415,10 @@ be dynamically uniform within the 'Scope' of the cooperative matrix type.
...
+elements of the coresponding matrix operands are sign-extended to the
+precision of 'Result Type', otherwise they are zero-extended.
```
wooyoungqcom, line 420: Don't we need to include MatrixResultSignedComponentsKHR  ?
I think when A has "MatrixASignedComponentsKHR" and the result does not have MatrixResultSignedComponentsKHR, we want to zero-extend the element of A.

extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
jeffbolznv: I copied this from somewhere else, I'll fix all the instances of it.

extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
jeffbolznv: When A has MatrixASignedComponentsKHR, it should be sign-extended regardless of the result type.

wooyoungqcom: Ah, Right. That seems consistent with the ANSI C integer promotion rules : "$6.3.1.1 ... if an int can represent all
values of the original type (as restricted by the width, for a bit-field), the value is converted to an
int; otherwise, it is converted to an unsigned int. ... The integer promotions preserve value including sign."

bashbaug:
Merging as discussed in the August 14th teleconference.

PR #265, opened 2024-08-07
State that in DebugLine the end column can be equal to the start column

Naghasan:
GLSL can in some cases generate this pattern.

Note: the document was regenerated using asciidoctor instead of asciidoc, hence the fixes in 3 tables.

### DIFF
--- a/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
+++ b/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
@@ -38,7 +38,7 @@ Contributors to original *OpenCL.DebugInfo.100* specification.
 Notice
 ------
 
-Copyright (c) 2019-2020 The Khronos Group Inc. Copyright terms at
+Copyright (c) 2019-2024 The Khronos Group Inc. Copyright terms at
 http://www.khronos.org/registry/speccopyright.html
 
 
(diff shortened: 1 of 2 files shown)


nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
```diff
@@ -1486,7 +1486,7 @@ instruction. +
  column number where the location begins. +
  +
  'Column End' is the '<id>' of a 32-bit integer *OpConstant* denoting the source
- column number where the location ends. This must be greater than 'Column Start'. +
+ column number where the location ends. This must be greater then or equal to 'Column Start'. +
```
dgkoch, line 1489:
```suggestion
 column number where the location ends. This must be greater than or equal to 'Column Start'. +
```

Naghasan: oh my ... thanks

dgkoch (approved):

alan-baker:
I notice you changed the generator from asciidoc to asciidoctor. That's not necessarily wrong, but it might be odd for people familiar with the previous versions of the spec. In my mind we should maintain the style as much as possible, but maybe it's not worth it.

bashbaug:
Merging as discussed in the August 21st teleconference.

PR #267, opened 2024-08-08
Add Github actions workflow to generate and check HTML specs

kpet:
Check that the HTML files in the repo are identical to those generated from the asciidoc files using the docker-* targets.

Change-Id: I4495cfc6504f3650118bf50fc37c96ab2f3bef82

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -1,0 +1,17 @@
+name: Presubmit
+on: [push, pull_request]
+
+jobs:
+  checkhtml:
+    name: Check generated HTML
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Generate HTML specs from asciidoc
+        run: |
+          touch extensions/*/*.asciidoc
+          make docker-all
+      - name: Check generated HTML matches HTML in repo
+        run: |
+          git diff --stat
+          git diff --exit-code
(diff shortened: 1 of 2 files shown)


.github/workflows/presubmit.yml
```diff
@@ -0,0 +1,17 @@
...
+jobs:
+  checkhtml:
+    name: Check generated HTML
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Generate HTML specs from asciidoc
+        run: |
+          touch extensions/*/*.asciidoc
+          make docker-all
```
Naghasan, line 13: can't we run the job in the container directly ?

dgkoch:
Won't this require everyone to use the *exact* same version of asciidoctor as is in CI?  That seems a bit annoying..

Naghasan:
The workflow uses the docker, so that should be easy to manage assuming we want contributors to use docker as well

dgkoch:
> It would but, with the work @Naghasan has done in #247, it's become trivial to use the official docker images. I think the benefits of guaranteeing correctness of the sources and that the HTML files do correspond to them and all use the same toolchain are worth the IMO minor inconvenience that is the forced used of docker. Happy to discuss this further in a future teleconference.

It's not about how easy it is to use docker. There are non-technical concerns..
https://www.docker.com/pricing/

kpet:
2024/08/14 teleconference: consensus in the group to skip this as a step and transition to automatically publishing HTML specs via Github Pages and remove HTML files from the repo.

PR #268, opened 2024-08-08
Disable asciidoctor footer and enable failure on errors

kpet:
- Disable footer with automatically generated "Last modified time" so HTML files don't change when the asciidoc source does not. This makes landing #267 much easier.
- Fail HTML file generation when asciidoctor encounters an error. This makes it easier for contributors to spot and fix errors.

Change-Id: I719718c40a8dd09c1109484b2c31c9c16f75ee90


Naghasan (approved):

bashbaug (approved):

bashbaug:
Merging as discussed in the August 14th teleconference.

PR #271, opened 2024-08-14
Add HTML building action

llvm-beanz:
This adds a basic GitHub action to build the HTML documentation using asciidoctor and upload the generated HTML artifact.

This build fails right now due to errors from asciidoctor, but if it's useful it is a good way to see what is going on.

### DIFF
--- a/.github/workflows/html-build.yml
+++ b/.github/workflows/html-build.yml
@@ -14,14 +14,13 @@ jobs:
   # Build job
   build:
     runs-on: ubuntu-latest
+    # We now refer to the container by its SHA instead of the name, to prevent
+    # caching problems when updating the image.
+    # container: khronosgroup/docker-images:asciidoctor-spec.20240726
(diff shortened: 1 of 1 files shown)


.github/workflows/html-build.yml
```diff
@@ -0,0 +1,34 @@
...
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+      - name: Install dependencies
+        run: |
+          sudo apt-get update
+          sudo apt -y install dos2unix
+          sudo apt -y install asciidoctor
```
Naghasan, line 24: Might be good to use the docker image (as suggested in the build docs and as used in Vulkan-Docs)

```suggestion
    # We now refer to the container by its SHA instead of the name, to prevent
    # caching problems when updating the image.
    # container: khronosgroup/docker-images:asciidoctor-spec.20240726
    container: khronosgroup/docker-images@sha256:089687083ceb36483a3917389e4278718ab19c594099634f5dd80e22540c960f
    steps:
      - name: Checkout
        uses: actions/checkout@v3
```

PR #272, opened 2024-08-20
Regenerate KHR extensions using asciidoctor

Naghasan:
The patch regenerated all extension using asciidoctor, it fixes issues encountered in the process: `cols` specifiers, extra or missing columns, odds behaviours of asciidoctor and directives ignored by asciidoc but honoured by asciidoctor.


bashbaug:
Discussed in the August 21st teleconference.  Need to check that the generated HTML all looks good.

bashbaug:
I took a quick look through the generated HTML output:

* SPV_KHR_float_controls2: this has the issue where the `'single quotes'` markup is not getting interpreted as italics.
* SPV_KHR_physical_storage_buffer: this one has some odd formatting with quoted asciidoc text vs. rendered text, though the odd formatting is in the previous asciidoc HTML output, also.  Suggest creating a separate issue to revisit.
* SPV_KHR_shader_atomic_counter_ops: this has some odd formatting also, though not as much as SPV_KHR_physical_storage_buffer.
* SPV_KHR_uniform_group_instructions: this has the single quote issue also.

I didn't find any other obvious formatting issues, and none of these are catastrophic.  I'd be fine merging this as-is and addressing these comments in a subsequent PR, if desired.

bashbaug:
I created a separate issue for SPV_KHR_physical_storage_buffer and SPV_KHR_shader_atomic_counter_ops, which have the quoted asciidoc source.

If we're OK with the single quote behavior in SPV_KHR_float_controls2 and SPV_KHR_uniform_group_instructions we can merge this as-is, though I would suggest creating an issue so we don't lose track of it also.

bashbaug:
Merging as discussed in the September 25th teleconference.

PR #276, opened 2024-08-26
Regenerate Google extensions using asciidoctor

alan-baker:
Please compare the rendered HTML (especially the tables) to the current content.


dneto0 (approved):
LGTM.

The first two have the same content as the HTML on 'main'.

For SPV_GOOGLE_user_type, the new HTML content matches the .asciidoc file on main. But it turns out the HTML on main is out of date.  
This patch brings everything back into sync.

thanks!

bashbaug:
Merging as discussed in the August 28th teleconference.

PR #282, opened 2024-09-03
update the extension HTML toolchain to enable table striping

### LINKED ISSUE #277: consider: update extension HTML toolchain to add table striping
From a review comment in an asciidoc -> asciidoctor update PR: https://github.com/KhronosGroup/SPIRV-Registry/pull/274#issuecomment-2309905784

The core SPIR-V spec enables asciidoctor "table striping" (via boilerplate.asciidoc).
https://docs.asciidoctor.org/asciidoc/latest/tables/striping/

This is most apparent for multi-line tables, where every other row gets shaded light grey, but it also affects the tables describing SPIR-V instructions, which generally have two rows, causing the second operand row light grey.

Should we enable table striping for HTML extensions as well, for uniformity with the core SPIR-V spec?  We can do this via the Makefile (by passing `-a table-stripes=even`) so individual extensions would not need to change.

bashbaug:
Fixes #277 

Enables table striping via the extension Makefile. This improves consistency with the main SPIR-V spec and improves readability for long tables.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@ ASCIIDOC_HTML_OPTIONS = --trace --failure-level ERROR \
   -a nofooter \
   -a stylesdir=$(SPIRV_DIR)/resources \
   -a stylesheet=spirv.css \
+  -a table-stripes=even \
   -a toc2 \
   -a toclevels=1
 


kpet (approved):

bashbaug:
Merging as discussed in the September 4th teleconference.

PR #284, opened 2024-09-25
Fix typo in optnone extension and regen using docker TC

Naghasan:


bashbaug (approved):
Merging as discussed in the September 25th teleconference.

Issue #280, opened 2024-09-02
Clarify DWARF Version operand of DebugCompilationUnit


MrSidims:
On windows default debug info format is CodeView and we would like to have an ability to express CodeView version in SPIR-V. For [OpenCL.DebugInfo.100](https://registry.khronos.org/SPIR-V/specs/unified1/OpenCL.DebugInfo.100.html) there are not much alternatives but to set zero as `DWARF Version` literal operand of `DebugCompilationUnit`. Starting from [NonSemantic.Shader.DebugInfo.100](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc) there are potential alternatives.

Currently our SPIR-V producer ([LLVM IR to SPIR-V translator](https://github.com/KhronosGroup/SPIRV-LLVM-Translator)) generates zero constant as `DWARF Version` operand of `DebugCompilationUnit` for CodeView and it becomes hard for the SPIR-V consumer to decide on the debug info version to set.

The questions/proposals I have:
1. Can we modify `DebugCompilationUnit` within NonSemantic.Shader.DebugInfo.100 to allow `DWARF Version` to also be `OpString`? Like this we could encode not only DWARF, but any other potential debug info formats.
2. If the answer to the above proposal is "no", then what should be the value of DWARF Version for non-DWARF debug info formats? On the one hand the specification says, that `DWARF Version` must be a 32-bit integer constant (and so in the translator we generate OpConstant with zero value), on the other hand [DebugInfoNone](https://github.com/KhronosGroup/SPIRV-Registry/blob/main/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc#DebugInfoNone) "Other instructions can refer to this one in case the debugging information is unknown, not available, or not applicable.", which seems to be a perfect fit in this case.

MrSidims:
@baldurk hi, may I ask you to express your opinion on that?

baldurk:
This was ported directly from the OpenCL spec, as I prioritised making as few breaking changes as possible since the original goal was to make some code that generated debug info usable on vulkan. I don't really understand the purpose of this field or what SPIR-V tools could feasibly use it for, but we definitely can't make backwards-compatible breaking changes.

[...]

PR #287, opened 2024-10-07
Allow DebugTypePointer BaseType argument substitution by DebugInfoNone

bwlodarcz:
This update addresses the DebugTypePointer instruction, which represents pointer types in the SPIR-V NonSemantic.Shader.DebugInfo.100 standard. Currently, the instruction requires a defined Base Type argument to comply with the standard. This requirement complicates the implementation of pointers to abstract types, such as void * in C, which represents type erasure, since these types do not have an available or defined Base Type.

To accommodate such scenarios, the standard includes the DebugInfoNone instruction. This amendment allows for the substitution of the concrete Base Type argument in DebugTypePointer with DebugInfoNone, thereby simplifying the implementation process for abstract pointer types.

### DIFF
--- a/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
+++ b/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
@@ -614,7 +614,8 @@ Describe a pointer or reference data type. +
 {result_type}  +
  +
 'Base Type' is the '<id>' of a debugging instruction that represents the pointee
- type. +
+ type. If the 'Base Type' is not defined or not available in source program,
+ this operand must refer to <<DebugInfoNone,*DebugInfoNone*>>. + 
  +
 'Storage Class' is a 32-bit integer *OpConstant* containing the class of the memory where
  the object pointed to is allocated.  Possible values of this operand are described in the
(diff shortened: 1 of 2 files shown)


nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
```diff
@@ -614,7 +614,8 @@ Describe a pointer or reference data type. +
 {result_type}  +
  +
 'Base Type' is the '<id>' of a debugging instruction that represents the pointee
- type. +
+ type. If the 'Base Type' is not defined or not available in source program,
+ this operand must refer to <<DebugInfoNone,*DebugInfoNone*>>. +
```
Naghasan, line 618: I understand that's your picked solution that is also implemented in the SPIR-V backend but I'd like to understand how this is preferable to the `DebugTypeBasic` trick as suggested in the thread https://github.com/KhronosGroup/SPIRV-Registry/issues/280#issuecomment-2393436571

I think DebugTypeBasic with `FlagUnknownPhysicalLayout` could be a suitable solution.

bwlodarcz: @Naghasan I believe that answer https://github.com/KhronosGroup/SPIRV-Registry/issues/280#issuecomment-2393712365 addresses your question. I'm open to further discussion.

PR #288, opened 2024-10-08
Fix NonSemantic Shader DebugInfo parent name

spencer-lunarg:
The `NonSemantic.Shader.DebugInfo.100` keeps referencing a `Scope <Id>` but the [grammar](https://github.com/KhronosGroup/SPIRV-Headers/blob/main/include/spirv/unified1/extinst.nonsemantic.shader.debuginfo.100.grammar.json) only has `Scope <Id>` at `DebugScope` and `DebugInlinedAt` ... the rest use `Parent <Id>`

This was very confusing for tooling when `"Scope"` was not matching with the auto-generated grammar file

 (this is also wrong in `OpenCL.DebugInfo.100` which I assume where the mistake was copied from)

### DIFF
--- a/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
+++ b/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
@@ -34,6 +34,7 @@ Contributors to original *OpenCL.DebugInfo.100* specification.
  - Neil Henning, Codeplay
  - Kerch Holt, Nvidia
  - Jaebaek Seo, Google
+ - Spencer Fricke, LunarG
 
 Notice
 ------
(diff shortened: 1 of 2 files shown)


Naghasan (approved):
Changes looks ok on their own. However, `Scope` is used to describe the lexical scope so this makes me wonder if the issue is the grammar and not the text itself. @baldurk do you know ?

Approving as I think the change is ok and it is better than changing the grammar file.

bashbaug:
Merging as discussed in the October 16th teleconference.

spencer-lunarg:
@bashbaug I noticed the same issue in `OpenCL.DebugInfo.100` but not sure where that spec lives actually

Issue #270, opened 2024-08-08
Check HTML spec generation and asciidoc validity in CI


kpet:
The HTML files for the various extensions currently in the repository were generated with various versions of asciidoc or asciidoctor and we do not currently check that the HTML files in the repository match the corresponding asciidoc source.

Following the introduction of an official toolchain in https://github.com/KhronosGroup/SPIRV-Registry/pull/247, we should add automation to guarantee that:
- The HTML files correspond to the asciidoc source and have been generated with the official toolchain
- All asciidoc sources are free from syntax errors (many of the files currently in the repo contain errors)

I think the following sequence of steps should get us there:
- [ ] Remove footer in HTML files (#268). The footer contains the generation time and makes it impossible to diff in repo HTML files with files generated from the asciidoc in automation.
- [ ] Fix errors in asciidoc sources and update all HTML files in the repository to use the official asciidoctor toolchain
  - [ ] AMD: TBD
  - [ ] ARM: #269 
  - [ ] EXT: TBD
  - [ ] GOOGLE: TBD
  - [ ] HUAWEI: TBD
  - [ ] INTEL: TBD
  - [ ] KHR: TBD
  - [ ] NV: TBD
  - [ ] QCOM: TBD
- [ ] Add automation to guarantee the correctness of asciidoc sources and that in repo HTML files match the asciidoc source and were generated with the official toolchain (#267)

kpet:
2024/08/14 teleconference: consensus in the group to skip automatic checking of the HTML files as a step and transition directly to automatically publishing HTML specs via Github Pages and remove HTML files from the repo.

bashbaug:
In case it helps other vendors updating their extension specs: One of the differences I noticed updating the Intel specs was that text in `'single quotes'` is rendered as _italics_ using asciidoc, but not always using asciidoctor.  I couldn't figure out how to make the single quote method work reliably, and it's not documented in the asciidoctor docs, so I ended up switching everything to use `_underscores_` for italics instead.

wooyoungqcom:
SPV_EXT_mesh_shader.asciidoc seems broken. Quite a few tables are not correctly rendered when opened.

Tobski:
I've just raised https://github.com/KhronosGroup/SPIRV-Registry/pull/290 for EXT and AMD extensions

Tobski:
> Thanks so much, Tobias I browsed the generated html files, they all look OK, now.
> 
[...]

PR #289, opened 2024-10-13
Update generated HTML for huawei extensions

PanGao-h:
With the update of asciidoctor toolchains mentioned in https://github.com/KhronosGroup/SPIRV-Registry/issues/270 , the htmls files are regenerated. 

+ Change the suffix from _adoc_ to _asciidoc_
+ Fix some incorrent table syntax


Naghasan (approved):
Thanks for the update

bashbaug (approved):
LGTM!

PR #295, opened 2024-10-23
Fix build failure for QCOM extensions

dgkoch:
Includes fixes from #294 (to be able to test CI)

PR #296, opened 2024-10-23
fix automated builds

bashbaug:
* Updates the GitHub actions to build from the proper branch for pull requests
* Includes the asciidoc fixes from #295 to fix the asciidoc build errors

### DIFF
--- a/.github/workflows/html-build.yml
+++ b/.github/workflows/html-build.yml
@@ -1,14 +1,17 @@
 name: Rebuild all with asciidoctor
 
-on:
-  # Runs on pushes targeting the main branch.
-  push:
-    branches: ["main"]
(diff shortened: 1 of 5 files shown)


dgkoch (approved):
Sees to work, so LGTM. I don't really understand which piece fixes it though.

bashbaug:
I think it's due to the change from `pull_request_target` to `pull_request`.  See:
https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target

"This event runs in the context of the base of the pull request, rather than in the context of the merge commit, as the pull_request event does."

Naghasan (approved):
LGTM

PR #299, opened 2024-11-18
Regenerate non semantic extensions and add to default build

Naghasan:


bashbaug (approved):
Merging as discussed in the November 20th teleconference.

PR #301, opened 2024-12-03
Set actual date rather than using {docdate} and unify version format

Naghasan:
{docdate} will print the generation date, which will change with every PR merged. Change to actual dates.

Some documents used docdate for the Revision History, in which case date set is the PR date.

Address point 1 of https://github.com/KhronosGroup/SPIRV-Registry/issues/270#issuecomment-2471674678


bashbaug (approved):
Thanks for doing this!

I didn't check all of the dates, but I did a few spot checks, and everything looked correct to me.

bashbaug:
Merging as discussed in the December 4th teleconference.

PR #303, opened 2024-12-16
Relax DebugLine validation in NonSemantic.Shader.DebugInfo.100

### LINKED ISSUE #302: Incorrect validation rule in NonSemantic.Shader.DebugInfo.100 DebugLine
See [DebugLine](https://github.khronos.org/SPIRV-Registry/nonsemantic/NonSemantic.Shader.DebugInfo.100.html#DebugLine).

>Line Start is the <id> of a 32-bit integer OpConstant denoting the source line number where the location begins.

>Line End is the <id> of a 32-bit integer OpConstant denoting the source line number where the location ends. This may be the same as Line Start if the location doesn’t cover multiple lines.

>Column Start is the <id> of a 32-bit integer OpConstant denoting the source column number where the location begins.

>Column End is the <id> of a 32-bit integer OpConstant denoting the source column number where the location ends. This must be greater than or equal to Column Start.

The validation for `Column End` should only apply if `Line Start` equals `Line End`.  https://github.com/KhronosGroup/SPIRV-Tools/pull/5846 implemented the rule as is, but that causes some multi-line uses to be considered invalid.

alan-baker:
Fixes #302

* Column End >= Column Start only if Line End == Line Start

### DIFF
--- a/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
+++ b/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
@@ -1480,14 +1480,14 @@ instruction. +
  line number where the location begins. +
  +
  'Line End' is the '<id>' of a 32-bit integer *OpConstant* denoting the source
- line number where the location ends. This may be the same as 'Line Start' if
- the location doesn't cover multiple lines. +
+ line number where the location ends. This must be greater than or equal to 'Line End'. +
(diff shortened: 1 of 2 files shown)


Naghasan (approved):
LGTM, thanks

bashbaug:
Merging as discussed in the December 18th teleconference.

PR #312, opened 2025-02-04
Clarify NonSemantic Shader DebugInfo Line/Column

spencer-lunarg:
Every text editor I am aware of use 1-index for line/columns and I would imagine the same for Shader DebugInfo

This adds a section to clarify that a Line/Column of 1 is the start

### DIFF
--- a/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
+++ b/nonsemantic/NonSemantic.Shader.DebugInfo.100.asciidoc
@@ -55,8 +55,8 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2024-10-08
-| Revision           | 11
+| Last Modified Date | 2025-02-04
+| Revision           | 12
 |========================================
 
 Dependencies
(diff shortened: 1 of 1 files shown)


bashbaug (commented):
I wonder if we should match [DWARF5](https://dwarfstd.org/doc/DWARF5.pdf) even more closely.  From Section 6.2.2 of the DWARF5 specification:

> Line: An unsigned integer indicating a source line number. Lines are numbered beginning at 1. The compiler may emit the value 0 in cases where an instruction cannot be attributed to any source line.

> Column: An unsigned integer indicating a column number within a source line. Columns are numbered beginning at 1. The value 0 is reserved to indicate that a statement begins at the “left edge” of the line.

spencer-lunarg:
> I wonder if we should match [DWARF5](https://dwarfstd.org/doc/DWARF5.pdf) even more closely. From Section 6.2.2 of the DWARF5 specification:

Wow, that is WAY better of a description and basically want I wanted to say

bashbaug:
Hi @spencer-lunarg , What do you want to do with this PR?  Do you want to update it to align closer to the DWARF5 specification?

spencer-lunarg:
@bashbaug updated to match the DWARF5 spec almost word for word

Naghasan (approved):

bashbaug (approved):
Thanks!  LGTM.

bashbaug:
Merging as discussed in the March 5th teleconference.

PR #332, opened 2025-04-09
Move all Intel FPGA-specific extensions to new Altera folder.

aejjehint:
As agreed upon with Intel, we are moving all FPGA-specific extensions to a new ALTERA folder due to the separation of Altera from Intel. We will be keeping the names of the extensions unmodified for now, with plans to change them in the future.


Naghasan:
Thanks for the PR.

The folder the extension is in affects the URL so, as it is, we will possibly end up with dead links. Also the HTMLs are simple redirections now, so they'll point to the wrong place.

I'd suggest the following to avoid this:

- move the asciidoc into the `ALTERA` folder (as done here)
- change occurrences of `INTEL/` to `ALTERA/` in the HTMLs
- **copy** (or simlink) the HTMLs into the the `ALTERA` folder (not sure the symlink would work though)
- for extensions without their HTML in the repo, create one that redirect to the new location

The README also need to be updated.

PawelJurek (approved):

bashbaug (approved):

bashbaug:
Merging offline, as discussed in the April 16th teleconference.

PR #335, opened 2025-04-10
Specify which cooperative matrix instructions are tangled

jeffbolznv:
Stacked on https://github.com/KhronosGroup/SPIRV-Registry/pull/334.

### DIFF
--- a/extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
+++ b/extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
@@ -56,7 +56,7 @@ Version
 [width="40%",cols="25,25"]
 |========================================
 | Last Modified Date | 2025-04-10
-| Revision           | 8
+| Revision           | 9
 |========================================
 
 Dependencies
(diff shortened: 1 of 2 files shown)


extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
```diff
@@ -115,6 +115,11 @@ is a composite with an implementation-dependent number of components
 composite for all operations that act on composite types. The mapping
 of components to invocations and indexes is implementation-dependent.
 
+Add to the list of Tangled Instructions in section 2.2.5 Control Flow:
+
+* *OpCooperativeMatrixLoadKHR*, *OpCooperativeMatrixStoreKHR*,
+  *OpCooperativeMatrixMulAddKHR*, and type conversion instructions when the input is a cooperative matrix.
```
bashbaug, line 121: I'm a little surprised to see the type conversion instructions in this list, but not the arithmetic instructions.  Assuming this is correct, for my own education, what is a good mental model to understand this difference?

jeffbolznv: For the conversion instructions, multiple implementations (including NVIDIA's) have different mappings of (row,col)->invocation for different types, so conversions may need to move data between invocations and thus need to be tangled.

I'm on the fence about whether arithmetic instructions should be tangled. I don't know of a practical reason that they need to be tangled, but I also don't know of anything useful you could do with them in divergent control flow. And from a spec purism point of view it may make more sense for them to be tangled.

bashbaug: Yeah, I think it's an easier mental model if the arithmetic instructions were considered to be tangled, also.

How about OpCompositeConstruct?  Does it make sense to construct a cooperative matrix in divergent control flow?

jeffbolznv: OpCompositeConstruct is similar to the arithmetic ops. I can't think of something useful you can do with this, but it's also not obviously harmful.

jeffbolznv: In the telecon we agreed to add arithmetic/compositeconstruct. I'll update the change soon.

bashbaug (approved):

Naghasan (approved):

bashbaug:
Merging as discussed in the April 23rd teleconference.

PR #341, opened 2025-05-02
add ratification dates to specifications

bashbaug:
These two EXT extensions were recently ratified:
* SPV_EXT_mesh_shader
* SPV_EXT_opacity_micromap

This KHR extension was previously ratified, but did not include the ratification date in the specification:
* SPV_KHR_8bit_storage


dgkoch (approved):
LGTM

PR #346, opened 2025-06-11
Add SPV_ARM_graph specification

kpet:
Change-Id: I403050fe790bd0590a28ef6c3a2a7027049d5efa

### DIFF
--- a/extensions/ARM/SPV_ARM_graph.asciidoc
+++ b/extensions/ARM/SPV_ARM_graph.asciidoc
@@ -33,7 +33,7 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2025-06-11
+| Last Modified Date | 2025-06-18
 | Revision           | 1
 |========================================
 
@@ -43,7 +43,7 @@ Dependencies
 This extension is written against the SPIR-V Specification,
 Version 1.6 Revision 2.
 
-This extension requires SPIR-V 1.6.
+This extension requires SPIR-V 1.0.
 
 This extension requires SPV_ARM_tensors.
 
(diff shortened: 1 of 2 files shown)


extensions/ARM/SPV_ARM_graph.asciidoc
```diff
@@ -0,0 +1,345 @@
...
+| Revision           | 1
+|========================================
+
+Dependencies
+------------
+
+This extension is written against the SPIR-V Specification,
+Version 1.6 Revision 2.
+
+This extension requires SPIR-V 1.6.
```
Naghasan, line 46:
```suggestion
This extension requires SPIR-V 1.0.
```

nit, I don't see anything that would require 1.6

extensions/ARM/SPV_ARM_graph.asciidoc
```diff
@@ -0,0 +1,345 @@
...
+
+'Graph Interface Type': An *OpTypeTensorARM*, *OpTypeArray* of *OpTypeTensorARM*, or
+*OpTypeRuntimeArray* of *OpTypeTensorARM*.
+
+Logical Layout of a Module
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+This extension changes the logical layout of modules as follows:
+
+- Add a new section after 11 Function definitions where graph definitions and entry points reside.
```
Naghasan, line 88: Didn't you mean to define graph definition here ? there is no defs attached here (only the last 2 points of the validation rules)

extensions/ARM/SPV_ARM_graph.asciidoc
kpet: I can't think of anything either. No objection to lowering the requirement to 1.0.

extensions/ARM/SPV_ARM_graph.asciidoc
kpet: Yes, better to expand on it here as is done for functions in the core spec. Thanks!

PR #348, opened 2025-06-18
SPV_ARM_tensors: allow NonReadable/NonWritable decorations

kpet:
on tensor variables in the UniformConstant storage class.

Change-Id: I5146a29bf83266ba99ff68e2b6fe9494600fd48d

### DIFF
--- a/extensions/ARM/SPV_ARM_tensors.asciidoc
+++ b/extensions/ARM/SPV_ARM_tensors.asciidoc
@@ -93,7 +93,7 @@ Modify Section 3.20, "Decoration", adding
 
 "A tensor variable in the *UniformConstant* storage class (see **OpTypeTensorARM**)"
 
-to the list of types for which a memory object declaration may be decorated with
+to the list of cases in which a memory object declaration may be decorated with
 *NonWritable* or *NonReadable*.
 
 Capability


extensions/ARM/SPV_ARM_tensors.asciidoc
```diff
@@ -86,8 +86,18 @@ Modify the definition of the following terms in 2.2.2 Types:
...
-Capabilities
-~~~~~~~~~~~~
+Decoration
+~~~~~~~~~~
+
+Modify Section 3.20, "Decoration", adding
+
+"A tensor variable in the *UniformConstant* storage class (see **OpTypeTensorARM**)"
+
+to the list of types for which a memory object declaration may be decorated with
```
kpet, line 96:
```suggestion
to the list of cases in which a memory object declaration may be decorated with
```

extensions/ARM/SPV_ARM_tensors.asciidoc
```diff
@@ -86,8 +86,18 @@ Modify the definition of the following terms in 2.2.2 Types:
...
-~~~~~~~~~~~~
+Decoration
+~~~~~~~~~~
+
+Modify Section 3.20, "Decoration", adding
+
+"A tensor variable in the *UniformConstant* storage class (see **OpTypeTensorARM**)"
+
+to the list of cases in which a memory object declaration may be decorated with
+*NonWritable* or *NonReadable*.
```
bashbaug, line 97: Sorry for the potentially silly question - I feel like I must be missing something here.

If the tensor variable is in the **UniformConstant** storage class isn't it already read-only (and hence **NonWritable**?  Additionally, what would it mean to have a read-only variable that is **NonReadable**?

kpet: Similarly to what happens for images, the variable in the **UniformConstant** storage class just is a handle to the tensor data. The handle itself is always read-only but the data it points to can be read-write, read-only, or write-only. I agree it is a little bit confusing. You can have a look at the validator changes in https://github.com/KhronosGroup/SPIRV-Tools/pull/6184 to observe the symmetry with images.

bashbaug: Got it - thanks!

bashbaug (approved):

bashbaug:
Merging as discussed in the July 2nd teleconference.

PR #354, opened 2025-09-02
Update logical pointer validation rules for SPV_NV_cluster_acceleration_structure and SPV_NV_linear_swept_spheres.

dnovillo:
See SPIR-V issue 878.

### DIFF
--- a/extensions/NV/SPV_NV_cluster_acceleration_structure.asciidoc
+++ b/extensions/NV/SPV_NV_cluster_acceleration_structure.asciidoc
@@ -31,8 +31,8 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2025-09-03
-| Revision           | 3
+| Last Modified Date | 2025-09-02
+| Revision           | 2
 |========================================
 
 Dependencies
@@ -90,7 +90,7 @@ New Instructions
 Instructions added under the *RayTracingClusterAccelerationStructureNV* capability
 
 ----
-OpRayQueryGetClusterIdNV
+OpRayQueryGetIntersectionClusterIdNV
 OpHitObjectGetClusterIdNV
 ----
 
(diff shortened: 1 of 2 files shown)


extensions/NV/SPV_NV_cluster_acceleration_structure.asciidoc
```diff
@@ -185,6 +186,20 @@ Allowed in *AnyHitKHR* and *ClosestHitKHR* ray tracing execution models.
...
++
+====
+It is invalid for a pointer to be an operand to any instruction other than:
+====
++
+to include:
++
+====
+* <<OpHitObjectGetClusterIdNV,*OpHitObjectGetClusterIdNV*>>
+* <<OpRayQueryGetIntersectionClusterIdNV,*OpRayQueryGetIntersectionClusterIdNV*>>
```
alan-baker, line 201: spirv.core.grammar.json has this named `OpRayQueryGetClusterIdNV`. Probably worth updating to match.

alan-baker (commented):
Changes look good, just the minor pre-existing issue in naming.

dnovillo:
> Changes look good, just the minor pre-existing issue in naming.

Actually, I've been chatting with the ray tracing folks. It seems that the grammar is wrong. The name should be `OpRayQueryGetIntersectionClusterIdNV `. I'll update the grammar as well.

Naghasan:
>  I'll update the grammar as well.

Maybe best to update this using an alias (`aliases` entry) in order to avoid breaking changes

dnovillo:
> > > I'll update the grammar as well.
> > 
> > 
> > Maybe best to update this using an alias (`aliases` entry) in order to avoid breaking changes
> 
> Thanks. I'll do that.

This is now done.

bashbaug (approved):
Merging as discussed in the September 10th teleconference.

PR #355, opened 2025-09-02
Modify logical pointer validation rules (spir-v#878) for SPV_NV_ray_t…

pradyumans:
…racing_motion_blur.asciidoc for


pradyumans:
For SPV_NV_shader_invocation_reorder I've left out OpTypeHitObjectNV (since it's a type and doesn't handle ptr operands directly) and OpReorderThreadWithHintNV (since it takes generic hint and bits as 32bit integers and not pointers). All other instructions are covered.

PR #356, opened 2025-09-15
Publish first revision of TOSA.001000.1 extended instruction set specification

kpet:


extended/TOSA.001000.1.html
bashbaug, line 1: Checking: The intent is to commit the rendered HTML to this repo vs. the asciidoc source?

I'm not sure we have everything wired up to serve HTML committed to this repo.  Remember that for other specifications the HTML files are automatically generated from the asciidoc source and then uploaded to GitHub pages.

Naghasan (approved):

PR #358, opened 2025-09-22
SPV_KHR_cooperative_matrix: clarify requirements on Stride

### LINKED ISSUE #279: Unclear requirements on Stride operand to OpCooperativeMatrix{Load,Store}KHR
For the **RowMajorKHR** and **ColumnMajorKHR** memory layouts the SPV_KHR_cooperative_matrix specification currently states

>Stride must be greater than 0 when passed to OpCooperativeMatrixStoreKHR and must be greater than or equal to 0 when passed to OpCooperativeMatrixLoadKHR.

@alan-baker observed in https://github.com/KhronosGroup/SPIRV-Tools/pull/5777#pullrequestreview-2266303751 that the specification does not explicitly document the signedness of the _Stride_ operand.

This was discussed in the 2024/08/28 SPIR teleconference and the consensus seemed to be that _Stride_ is meant to always be interpreted as unsigned. If that's the case we should state it in the specification and remove the requirement that it be "greater than or equal to 0 when passed to **OpCooperativeMatrixLoadKHR**".

Also, we observed in the teleconference that the requirement for _Stride_ to be "greater than 0 when passed to **OpCooperativeMatrixStoreKHR**" was necessary for correctness but not sufficient. Should we strengthen the rule and require that the stride be greater than the total size of the data being stored to avoid write collisions?

kpet:
Fixes #279

### DIFF
--- a/extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
+++ b/extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
@@ -55,8 +55,8 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2025-04-24
-| Revision           | 10
+| Last Modified Date | 2025-09-22
+| Revision           | 11
 |========================================
 
 Dependencies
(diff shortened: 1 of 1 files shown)


extensions/KHR/SPV_KHR_cooperative_matrix.asciidoc
```diff
@@ -227,9 +227,9 @@ are laid out with a fixed stride communicated via the 'Stride' operand to
 provided. Elements (row,$$*$$) of the result of load or store operations are taken
 in order from contiguous locations starting at 'Pointer'[row*'Stride'] where
 'Pointer' is the 'Pointer' operand to *OpCooperativeMatrixLoadKHR* or
-*OpCooperativeMatrixStoreKHR*. 'Stride' must be greater than 0 when passed to
-*OpCooperativeMatrixStoreKHR* and must be greater than or equal to 0 when passed
-to *OpCooperativeMatrixLoadKHR*.
+*OpCooperativeMatrixStoreKHR*. 'Stride' is interpreted as unsigned and must be
+greater than 0 when passed to
+*OpCooperativeMatrixStoreKHR*.
```
kpet, line 232: We did discuss potentially requiring the stride be such that no write collisions occur. Do we want to do this too?

PR #364, opened 2025-10-20
rename Intel FPGA extensions to Altera

bashbaug:
This PR renames the FPGA-specific Intel extensions from Intel extensions with Intel enumerants to Altera extensions with Altera enumerants.  Please note that the values of the enumerants have not changed, only the names of the enumerants.  All of the affected extensions were already in the Altera directory in this repo.

To ensure there are no broken links, this PR also adds HTML redirects from the prior Intel extension to the new Altera extension spec, though only the new Altera extensions are linked from the README.


bashbaug:
Hi @intel-jisheng1 , thanks for the quick review!

Could I suggest keeping this PR focused on the extensions already in the Altera folder?  If we decide that other extensions should additionally be considered, such as the SPV_INTEL_kernel_attributes or SPV_INTEL_global_variable_host_access extensions that you've identified above, we could do a separate PR that moves the file and does the renaming similar to what I've done in this PR.

PR #347, opened 2025-06-11
Replace "Enabling Capabilities" with "Implictly Declares" in capability tables

kpet:

PR #366, opened 2025-10-28
fix SPV_EXT_shader_64bit_indexing capability table

bashbaug:
Capabilities "implicitly declare" other capabilities rather than being "enabled by" other capabilities.


bashbaug:
Tagging @jeffbolznv.  Sorry I didn't notice this previously!

See also #347 (and #345).

PR #370, opened 2025-11-12
Clarify naming conventions for opaque types in NSDI.100

dnovillo:
This stops the NSDI.100 spec from mandating a specific mangling for opaque types. It updates 'Name' requirements to indicate it should be mangled in an implementation-defined manner to avoid clashes with user-defined names.

This should help address issues like https://github.com/KhronosGroup/glslang/issues/4063


bashbaug (approved):
Merging as discussed in the November 19th teleconference.

PR #395, opened 2026-03-30
Fix binary form layout for NonSemantic instructions

spencer-lunarg:
Currently we have

image

which is wrong, the disassembled SPIR-V looks like

```
%14 = OpExtInst %void %13 1 %11 %12
```

So this fixes it in `NonSemantic.DebugPrintf` and `NonSemantic.DebugBreak`

(also it generates the HTML files which seemed to have been broken anyway)

now it looks like

image

(Also there should be no need for the capability, going off how `NonSemantic.Shader.DebugInfo` is doing it as well)

### DIFF
--- a/nonsemantic/NonSemantic.DebugBreak.asciidoc
+++ b/nonsemantic/NonSemantic.DebugBreak.asciidoc
@@ -16,11 +16,12 @@ Contributors
 - Qingyuan Zheng, NVIDIA Corporation
 - Ashwin Lele, NVIDIA Corporation
 - Jeff Bolz, NVIDIA Corporation
+- Spencer Fricke, LunarG Inc
 
 Notice
 ------
 
-Copyright (c) 2019-2022 The Khronos Group Inc. Copyright terms at
+Copyright (c) 2019-2026 The Khronos Group Inc. Copyright terms at
 http://www.khronos.org/registry/speccopyright.html
 
 Status
@@ -33,8 +34,8 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2022-08-18
-| Revision           | 1
+| Last Modified Date | 2026-04-13
+| Revision           | 2
 |========================================
 
 Dependencies
@@ -49,7 +50,7 @@ This instruction set requires SPIR-V 1.0 and SPV_KHR_non_semantic_info.
 ---------------
 
 This specifies the NonSemantic.DebugBreak extended instruction set. It
-provides a *DebugBreak* instruction which is a hint for any attached 
+provides a *DebugBreak* instruction which is a hint for any attached
 shader debugger to hit a breakpoint.
 
 
@@ -59,20 +60,16 @@ Import this extended instruction set using an OpExtInstImport
 2. Binary Form
 ---------------
 
-[cols="1,1,2*3",width="100%"]
+[cols="1",width="100%"]
 |=====
-3+|[[DebugBreak]]*DebugBreak* +
+1+|[[DebugBreak]]*DebugBreak* +
  +
 Hint the attached shader debugger to trigger a breakpoint and
 pause the execution. If no debugger is available, this instruction
 should be ignored. +
  +
 'Return Type' must be *OpTypeVoid*.
-1+|Capability:
-
-1+| 1 
-1+| 1
-2+|
+| 1
 |=====
 
 
@@ -90,4 +87,5 @@ Revision History
 |========================================
 |Rev|Date|Author|Changes
 |1|2022-08-18|Qingyuan Zheng|Initial revision
+|2|2026-04-13|Spencer Fricke|Fix binary form view
 |========================================
--- a/nonsemantic/NonSemantic.DebugBreak.asciidoc
+++ b/nonsemantic/NonSemantic.DebugBreak.asciidoc
@@ -60,7 +60,7 @@ Import this extended instruction set using an OpExtInstImport
 2. Binary Form
 ---------------
 
-[cols="1",width="100%"]
+[cols="1"]
 |=====
 1+|[[DebugBreak]]*DebugBreak* +
  +
--- a/nonsemantic/NonSemantic.DebugPrintf.asciidoc
+++ b/nonsemantic/NonSemantic.DebugPrintf.asciidoc
@@ -14,11 +14,12 @@ Contributors
 ------------
 
 - Jeff Bolz, NVIDIA Corporation
+- Spencer Fricke, LunarG Inc
 
 Notice
 ------
 
-Copyright (c) 2019-2020 The Khronos Group Inc. Copyright terms at
+Copyright (c) 2019-2026 The Khronos Group Inc. Copyright terms at
 http://www.khronos.org/registry/speccopyright.html
 
 Status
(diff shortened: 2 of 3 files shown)


bashbaug (changes requested):
Note, I'm not super familiar with these non-semantic extended instruction sets, so I didn't review the actual proposed changes.

spencer-lunarg:
Fixed/reverted the HTML

> I'm not super familiar with these non-semantic extended instruction sets

https://github.com/KhronosGroup/SPIRV-Guide/blob/main/chapters/nonsemantic.md (which I want to update after this change now knowing more)

also if you want to see an easy example in action https://godbolt.org/z/P9T75fjez

nonsemantic/NonSemantic.DebugPrintf.asciidoc
```diff
@@ -57,9 +58,9 @@ Import this extended instruction set using an OpExtInstImport
 2. Binary Form
 ---------------
 
-[cols="1,1,2*3",width="100%"]
+[cols="1,1,3*2,1,2*2",width="100%"]
 |=====
-3+|[[DebugPrintf]]*DebugPrintf* +
+8+|[[DebugPrintf]]*DebugPrintf* +
```
alan-baker: This is sort of a hybrid style between what I've seen in extended instruction sets. It is more like how glsl.std.450 is defined (https://registry.khronos.org/SPIR-V/specs/unified1/GLSL.std.450.html). Basically just denoting the extra operands to OpExtInst. The extra part here is the word count.

Other sets (e.g. [Shader.DebugInfo](https://github.khronos.org/SPIRV-Registry/nonsemantic/NonSemantic.Shader.DebugInfo.html)) instead replicate the whole OpExtInst set of operands.

This might come down to personal preference, so I'd suggest getting the authors to weigh in.

nonsemantic/NonSemantic.DebugBreak.asciidoc
```diff
@@ -59,20 +60,20 @@ Import this extended instruction set using an OpExtInstImport
 2. Binary Form
 ---------------
 
-[cols="1,1,2*3",width="100%"]
+[cols="2*1,3*2,1",width="100%"]
 |=====
-3+|[[DebugBreak]]*DebugBreak* +
+6+|[[DebugBreak]]*DebugBreak* +
```
alan-baker: I agree there seems to be a problem here (aside from what I mentioned in DebugPrintf). There is an extra `1` operand.

nonsemantic/NonSemantic.DebugPrintf.asciidoc
spencer-lunarg: > so I'd suggest getting the authors to weigh in.

cc @jeffbolznv 

> This might come down to personal preference

I would rather have "a single way to do it" and then I will be happy to just update it, but the fact DebugPrintf is different than DebugBreak which was different than Shader.DebugInfo... at that point if we really don't care about making them uniform, it just seems like a sloppy "do what ever you want, since if you know you know" for NonSemantic instructions

alan-baker: I prefer the glsl.std.450 style personally. It's what I used for [ClspvReflection](https://github.khronos.org/SPIRV-Registry/nonsemantic/NonSemantic.ClspvReflection.html)), but that's neither here nor there.

spencer-lunarg: I have zero preference, happy to make these (and ShaderDebug) like the glsl.std.450/ClspvReflection if everyone else is ok with that

alan-baker: I prefer the glsl.std.450 style personally. It's what I used for [ClspvReflection](https://github.khronos.org/SPIRV-Registry/nonsemantic/NonSemantic.ClspvReflection.html)), but that's neither here nor there.

bashbaug (commented):
The HTML artifacts are removed so my prior comments have been addressed - thanks!

Still need an approval regarding the content of the changes, though.

nonsemantic/NonSemantic.DebugPrintf.asciidoc
spencer-lunarg: WG 4/8/2026 - agreed to just make them all like glsl.std.450

nonsemantic/NonSemantic.DebugBreak.asciidoc
```diff
@@ -59,20 +60,16 @@ Import this extended instruction set using an OpExtInstImport
...
 pause the execution. If no debugger is available, this instruction
 should be ignored. +
  +
 'Return Type' must be *OpTypeVoid*.
-1+|Capability:
-
-1+| 1 
-1+| 1
-2+|
+| 1
```
spencer-lunarg, line 72: image

nonsemantic/NonSemantic.DebugPrintf.asciidoc
```diff
@@ -71,10 +72,9 @@ Interpretation of the format specifiers is specified by the client API. +
 'Return Type' must be *OpTypeVoid*. +
  +
 'Format' must be an *OpString*.
-1+|Capability:
-1+| 2+variable | 1 | '<id>' +
-'Format' | Optional
-'<id>', '<id>', ...
+| 1
+| '<id>' 'Format'
+| Optional '<id>', '<id>', ...
```
spencer-lunarg, line 77: image

nonsemantic/NonSemantic.Shader.DebugInfo.asciidoc
```diff
@@ -473,25 +473,24 @@ Instructions
 Missing Debugging Information
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-[cols="2*1,3*2,1,0*3"]
+[cols="1,0*3"]
```
dnovillo: nit: just `[cols="1"]` here? The `0*n` pattern happens in a couple other places.

nonsemantic/NonSemantic.DebugBreak.asciidoc
```diff
@@ -59,20 +60,16 @@ Import this extended instruction set using an OpExtInstImport
 2. Binary Form
 ---------------
 
-[cols="1,1,2*3",width="100%"]
+[cols="1",width="100%"]
```
dnovillo: Removing the `width` here will make the cell render at natural width.

dnovillo (changes requested):
Just a couple of nits. Thanks for the cleanup!

dnovillo (approved):

bashbaug (approved):
I was just reading through the NonSemantic.Shader.DebugInfo extended instruction set and confusing myself - "why does every instruction have extended instruction ID 12!?" - and I think if these changes were in place I would have been less confused.

I'm not super happy losing the column with the word count since it does seem valuable, but I don't have a great idea how to include it, so I vote to get these changes merged in their current form and we can always experiment with additional fixes in a subsequent PR.

Thanks for doing this!

dnovillo:
> I was just reading through the NonSemantic.Shader.DebugInfo extended instruction set and confusing myself - "why does every instruction have extended instruction ID 12!?" - and I think if these changes were in place I would have been less confused.
> 
> I'm not super happy losing the column with the word count since it does seem valuable, but I don't have a great idea how to include it, so I vote to get these changes merged in their current form and we can always experiment with additional fixes in a subsequent PR.
> 
> Thanks for doing this!

Agreed. I think for now consistency will improve the status quo. I'm merging this.

PR #399, opened 2026-04-16
Clarify packing of small elements into constant data

Tobski:
cc @spencer-lunarg


spencer-lunarg (approved):

bashbaug:
Merging as discussed in the April 29th teleconference.

PR #401, opened 2026-04-21
Add SPV_AMD_weak_linkage extension

MrSidims:


extensions/AMD/SPV_AMD_weak_linkage.asciidoc
```diff
@@ -0,0 +1,134 @@
...
+== Modifications to the SPIR-V Specification, Version 1.6
+
+=== Linkage Type
+
+In section 3.17 "Linkage Type" add the following row to the Linkage Type table:
+
+[cols="1,4,4",options="header",width="100%"]
+|====
+2+| Linkage Type| <<Capability,Enabling Capabilities>> |
+3 | *Weak* +
```
alan-baker: This is a bit unusual. I'd expect:

```suggestion
3 | *WeakAMD* +
```

extensions/AMD/SPV_AMD_weak_linkage.asciidoc
```diff
@@ -0,0 +1,134 @@
...
+Same as the *Export* linkage type, but a function or global variable with this
+linkage type may be replaced during linking by one with *Export* linkage from
+another module.
+If multiple modules provide a definition of the same symbol and at least one
+has *Export* linkage, the linker must select an *Export* definition and discard
+all *Weak* definitions of that symbol.
+If only *Weak* definitions exist for a given symbol, the linker selects one
+and discards the rest.
+| *Linkage*
```
alan-baker: Introducing this under an existing cap is also a little non-standard. Not necessarily a problem though.

extensions/AMD/SPV_AMD_weak_linkage.asciidoc
```diff
@@ -0,0 +1,134 @@
...
+- Dmitry Sidorov, AMD
+
+== Notice
+
+Copyright (c) 2026 The Khronos Group Inc. Copyright terms at
+http://www.khronos.org/registry/speccopyright.html
+
+== Status
+
+- Draft
```
alan-baker: Still draft?

MrSidims: Thanks, will fix this!

extensions/AMD/SPV_AMD_weak_linkage.asciidoc
MrSidims: This comment and a comment about capability are fair, that is also something I would expect. But another [linkage type extension](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_linkonce_odr.html) has this style (no capability, no KHR postfix), so I decided to follow its pre-existing style. Renaming won't harm, so if we are fine with having a bit of discrepancy in names in Linkage Type table - I'll do it right away.

alan-baker: I think it makes sense to change the name.

MrSidims: Agreed, changed the name and added the capability

MrSidims:
Temporary moving to draft to avoid incidental merge - need to double check token assignment.

alan-baker (approved):

PR #405, opened 2026-04-29
Update SPV_AMD_gpu_shader_half_float.asciidoc

Tobski:
Fix version to match change log


bashbaug (approved):
Merging as discussed in the April 29th teleconference.

PR #406, opened 2026-05-01
Add NonSemantic.Graph.DebugInfo extended instruction set

kpet:


alan-baker:
Is there a reason to not fold this into the shader debug set as the next revision?

kpet:
As agreed in the 2026/05/13 SPIR teleconference, I've captured the rationale for this being a separate instruction set for now in an issue.

bashbaug (approved):
Merging as discussed in the May 20th teleconference.

PR #419, opened 2026-05-20
Fix word count in constant data issue

Tobski:
Existing count was based on an earlier draft with an extra operand.


bashbaug (approved):

bashbaug:
Merging as discussed in the May 20th teleconference.

PR #423, opened 2026-06-12
Clarify SPV_EXT vs SPV_KHR opacity_micromap behavior equivalence

dgkoch:
Spell out in the `SPV_KHR_opacity_micromap` Issues resolution that:
- Declaring either `OpExtension` string without `OpacityMicromapIdKHR` produces the same shader behavior.
- `SPV_EXT_opacity_micromap` alone does not imply OMMs are always present.
- Whether OMM data is consumed during traversal is gated by which Vulkan extension is enabled:
  - `VK_EXT_opacity_micromap`: OMMs always consumed when present.
  - `VK_KHR_opacity_micromap`: OMMs consumed only when the shader declares `OpacityMicromapIdKHR`.

Bumps Last Modified Date to 2026-06-12 and revision to 4.

Resolves the ambiguity raised in [vulkan/vulkan#4842](https://gitlab.khronos.org/vulkan/vulkan/-/issues/4842).

### DIFF
--- a/extensions/KHR/SPV_KHR_opacity_micromap.asciidoc
+++ b/extensions/KHR/SPV_KHR_opacity_micromap.asciidoc
@@ -153,25 +153,27 @@ Specifically:
 
 * Declaring an OpExtension string in a module does not, by itself,
   determine whether opacity micromap data is used during ray traversal.
-  That is gated by which Vulkan opacity_micromap feature is enabled at
-  the API level:
+  That is determined by which Vulkan opacity_micromap extension is
+  enabled at the API level:
 
-  ** With the VK_EXT_opacity_micromap `micromap` feature enabled,
-     opacity micromap data present in the acceleration structure is
-     always used during ray traversal.
+  ** With VK_EXT_opacity_micromap enabled, opacity micromap data
+     present in the acceleration structure is always used during ray
+     traversal.
 
   ** With the VK_KHR_opacity_micromap `micromap` feature enabled,
      opacity micromap data present in the acceleration structure is
      used during ray traversal only when the module declares the
-     OpacityMicromapIdKHR execution mode.
+     OpacityMicromapIdKHR execution mode with its
+     specialization-constant operand equal to VK_TRUE.
 
 * If ray traversal uses an acceleration structure that contains
   opacity micromap data with the VK_KHR_opacity_micromap `micromap`
   feature enabled but the module does not declare the
-  OpacityMicromapIdKHR execution mode, the traversal result is
-  undefined. This is not equivalent to skipping opacity micromaps;
-  applications must explicitly opt in via the execution mode when
-  they intend to use opacity micromap data.
+  OpacityMicromapIdKHR execution mode with its specialization-constant
+  operand equal to VK_TRUE, the traversal result is undefined. This is
+  not equivalent to skipping opacity micromaps; applications must
+  explicitly opt in via the execution mode when they intend to use
+  opacity micromap data.
 
 * Declaring the OpacityMicromapIdKHR execution mode requires the
   VK_KHR_opacity_micromap `micromap` feature to be enabled at the
--- a/extensions/KHR/SPV_KHR_opacity_micromap.asciidoc
+++ b/extensions/KHR/SPV_KHR_opacity_micromap.asciidoc
@@ -32,7 +32,7 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2026-06-12
+| Last Modified Date | 2026-07-07
 | Revision           | 4
 |========================================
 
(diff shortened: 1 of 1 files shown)


extensions/KHR/SPV_KHR_opacity_micromap.asciidoc
```diff
@@ -144,6 +144,17 @@ it adds a new execution mode named OpacityMicromapIdKHR. There is a strong
 desire to keep the existing behaviors when a shader does not use the new execution
 mode.
 
+Specifically, declaring either OpExtension string without the OpacityMicromapIdKHR
+execution mode produces the same shader behavior. The behavior of a shader that
+declares SPV_EXT_opacity_micromap is identical to that of a shader that declares
+SPV_KHR_opacity_micromap without OpacityMicromapIdKHR. Declaring
+SPV_EXT_opacity_micromap alone does not imply that opacity micromaps are always
+present; whether opacity micromap data is consumed during ray traversal is
```
dnovillo: Does "_present_" here refer to being present as a capability or in the AS (not familiar with this, so maybe that's why it's confusing to me).

dnovillo (changes requested):

dgkoch:
Iterated per RT TSG 2026-06-25 feedback (`600fec0d`). Three things changed:

1. **Reference the Vulkan opacity_micromap `micromap` feature being enabled at the API level**, rather than "declaring the extension string." This matches Vulkan's actual gating model (enabling the extension isn't enough — the feature has to be enabled too), and it disambiguates the KHR case where both an extension and a feature are in play.

2. **Restructured the resolution as a bulleted list** for readability.

3. **Explicit `undefined` for the mismatch case.** If a shader traces an acceleration structure that contains opacity micromap data with the `VK_KHR_opacity_micromap` `micromap` feature enabled but without declaring `OpacityMicromapIdKHR`, the traversal result is undefined — this is NOT equivalent to a skip-OMM ray flag. Applications must opt in via the execution mode when they intend to use opacity micromap data.

@HansKristian-Work @dnovillo — please take another look.

dnovillo (approved):
LGTM.

extensions/KHR/SPV_KHR_opacity_micromap.asciidoc
```diff
@@ -144,6 +144,35 @@ it adds a new execution mode named OpacityMicromapIdKHR. There is a strong
...
+     OpacityMicromapIdKHR execution mode.
+
+* If a shader traces against an acceleration structure that contains
+  opacity micromap data with the VK_KHR_opacity_micromap `micromap`
+  feature enabled but without declaring the OpacityMicromapIdKHR
+  execution mode, the traversal result is undefined. This is not
+  equivalent to skipping opacity micromaps; applications must
+  explicitly opt in via the execution mode when they intend to use
+  opacity micromap data.
+
```
wooyoungqcom, line 185: I suppose in order to use OMM either the VK_EXT_opacity_micromap or VK_KHR_opacity_micromap
`micromap` feature should be enabled.

With that, I can think of four different combinations.

1) With the VK_EXT_opacity_micromap `micromap` feature enabled,
1.1) A shader has SPV_EXT_opacity_micromap or SPV_KHR_opacity_micromap without OpacityMicromapIdKHR
1.2) SPV_KHR_opacity_micromap with OpacityMicromapIdKHR declared

2) With the VK_KHR_opacity_micromap `micromap` feature enabled
2.1) A shader has SPV_EXT_opacity_micromap or SPV_KHR_opacity_micromap without OpacityMicromapIdKHR
2.2) SPV_KHR_opacity_micromap with OpacityMicromapIdKHR declared

L159 covers 1.1. 
L163 covers 2.2.
L168 covers 2.1.

Does L159 also cover 1.2, meaning OpacityMicromapIdKHR is ignored?

mnetsch: Provided the KHR feature is also not enabled, 1.2 is illegal per the spec since OpacityMicromapIdKHR requires the RayTracingOpacityMicromapExecutionModeKHR capability (https://github.com/KhronosGroup/SPIRV-Registry/blob/main/extensions/KHR/SPV_KHR_opacity_micromap.asciidoc) and the capability is only valid in the vulkan client if the KHR feature is enabled (https://docs.vulkan.org/spec/latest/appendices/spirvenv.html#spirvenv-capabilities)

wooyoungqcom: Then, I suppose it would be better to state the case in this clarification.

dgkoch: Done (75f3b20) — added a fourth bullet stating the requirement explicitly.

dgkoch:
Propagating the same two corrections that landed on the parallel Vulkan-side MR (!8409) after Matt Netsch's review there:

- **VK_EXT_opacity_micromap is extension-gated, not feature-gated.** The EXT capability entry in `vk.xml` has been `<enable extension="VK_EXT_opacity_micromap"/>` since day one (2022), and `VUID-*-micromap-11636` gates on the extension being enabled. Only the KHR path is feature-gated. The EXT sub-bullet and the introductory line have been updated to reflect that (`822b0b2`).
- **The KHR opt-in requires the `OpacityMicromapIdKHR` operand to equal `VK_TRUE`.** Per VU 11636, declaring the execution mode with the operand set to `VK_FALSE` doesn't count as opting in. Reflected in the KHR sub-bullet and in the "traversal result is undefined" bullet (`822b0b2`).

The fourth bullet (added earlier for Wooyoung) is unchanged — it's about whether the module can declare the execution mode at all, which the operand value doesn't affect.

bashbaug:
Merging as discussed in the July 15th teleconference.

PR #431, opened 2026-06-29
SPV_KHR_fma: clarify that OpFmaKHR results may be decorated with FPFastMathMode

aobolensk:


aobolensk:
Related discussion: https://github.com/llvm/llvm-project/pull/206143#pullrequestreview-4591606797

@jmmartinez FYI

gnl21:
Thanks for making this change. I agree that this clarification is needed for OpenCL (it's actually not needed for Vulkan, which is why it was overlooked in the first place).

Rather than adding this into the specification of the instruction, I think we should add a section to the extension that explicitly changes the list of instructions for which FPFastMathMode is valid. Would you be able to update the PR with this? This would be something roughly like `In section 3.15 "FP Fast Math Mode", following "Only valid on …​" add FmaKHR`.

aobolensk:
> Thanks for making this change. I agree that this clarification is needed for OpenCL (it's actually not needed for Vulkan, which is why it was overlooked in the first place).
> 
> Rather than adding this into the specification of the instruction, I think we should add a section to the extension that explicitly changes the list of instructions for which FPFastMathMode is valid. Would you be able to update the PR with this? This would be something roughly like `In section 3.15 "FP Fast Math Mode", following "Only valid on …​" add FmaKHR`.

Thank you for the feedback, done. Please take a look

extensions/KHR/SPV_KHR_fma.asciidoc
```diff
@@ -98,6 +98,11 @@ _Operand 2_ | _<id>_ +
 _Operand 3_
 |=====
 
+=== FP Fast Math Mode
+
+In section 3.15 "FP Fast Math Mode", following "Only valid on ...", add
+*OpFmaKHR*.
+
```
gnl21: The sections that are changed are normally listed in order, so I think we should move this to the start of the changes (after "Modifications to the SPIR-V Specification..." and before "Capability"). Otherwise I think this PR is fine now. Thanks for the update.

aobolensk: Thanks for the feedback! applied remark

gnl21 (approved):

MrSidims (approved):

bashbaug:
Merging as discussed in the July 8th teleconference.

PR #434, opened 2026-07-02
Readability fix for SPV_EXT_ocp_microscaling_types

vmaksimo:
Styling fix to improve readability after rendering - attaching text block to the parent list item.


kpet (approved):
That reads better indeed. Thanks!

PR #438, opened 2026-07-10
SPV_KHR_ray_query: require Initialize + add missing AABB-candidate preconditions

dgkoch:
Fixes two related gaps in `SPV_KHR_ray_query` reported in the SPIR-V issue tracker:

- **[spir-v#947](https://gitlab.khronos.org/spirv/SPIR-V/-/issues/947)**: Instructions like `OpRayQueryGetRayFlagsKHR` and `OpRayQueryGetRayTMinKHR` don't state what happens if the ray query hasn't been initialized. `OpRayQueryProceedKHR` also doesn't require prior initialization. Add a single global statement at the start of the Ray Query Instructions section: any instruction other than `OpRayQueryInitializeKHR` has undefined behavior if `OpRayQueryInitializeKHR` was not previously executed on the same ray query object.

- **[spir-v#945](https://gitlab.khronos.org/spirv/SPIR-V/-/issues/945)**: `OpRayQueryGetIntersectionCandidateAABBOpaqueKHR` currently has no preconditions listed, unlike the analogous `OpRayQueryGetIntersectionFrontFaceKHR` which requires Proceed to have been called, the last Proceed to have returned true, and the current candidate to be of the appropriate type. Add matching preconditions requiring the current candidate to have Ray Query Candidate Intersection Type `RayQueryCandidateIntersectionAABBKHR`.

Cross-checked against the Vulkan spec ray_query chapter (`raytraversal.adoc`) and `GLSL_EXT_ray_query`: both are consistent with these being the intended preconditions and neither needs changes. DXR compatibility check: HLSL `CandidateProceduralPrimitiveNonOpaque()` is documented as valid only for AABB (procedural) candidates during a live `Proceed()`, and DXR treats method calls on uninitialized `RayQuery` as UB — aligned with the proposed rules.

Revision bumped to 17.

### DIFF
--- a/extensions/KHR/SPV_KHR_ray_query.asciidoc
+++ b/extensions/KHR/SPV_KHR_ray_query.asciidoc
@@ -39,8 +39,8 @@ Version
 
 [width="40%",cols="25,25"]
 |========================================
-| Last Modified Date | 2026-06-03
-| Revision           | 16
+| Last Modified Date | 2026-07-10
+| Revision           | 17
 |========================================
 
 Dependencies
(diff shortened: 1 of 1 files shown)


stu-s (approved):

bashbaug:
Merging as discussed in the July 22nd teleconference. (Ray Tracing TSG has signed off)

PR #442, opened 2026-07-29
SPV_QCOM_image_vector_read_write

wooyoungqcom:


bashbaug:
@wooyoungqcom note the asciidoctor build errors - can you fix these?

wooyoungqcom:
Do you know what asciidoctor version is used?
I have Asciidoctor 2.0.16 and I don't see any error locally

wooyoungqcom:
Fixed.

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+
+- Adam Brouwers-Harries, Qualcomm
+- Balaji Calidas, Qualcomm
+- Victor Lomuller, Qualcomm
+- Wooyoung Kim, Qualcomm
+
+Notice
+------
+
+Copyright (c) 2024 The Khronos Group Inc. Copyright terms at
```
alan-baker:
```suggestion
Copyright (c) 2026 The Khronos Group Inc. Copyright terms at
```

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+according to the specified access shape and
+returns them as a vector of four components.
+
+_Result Type_ is a vector type with four components. The element type is one of 32-bit *floating-point* type,
+16-bit *floating-point* type, and 32-bit signed/unsigned *integer* type.
+
+_Texture Sampled Image_ must be an object whose type is OpTypeSampledImage. The MS operand of the 
+underlying OpTypeImage must be 0.
+
+_Coordinate_ must be a vector of *floating-point* type, whose vector size is 2.
```
alan-baker, line 217: It says an OpTypeSampledImage (as long as it is not multisampled), but the coordinate restriction limits this to 2D or 1D arrayed images.

Is it intentional to not work with 3D or 2D arrayed images?

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+
+_Component Id_ must be a 32-bit *integer* type scalar and be one of 0, 1, 2, and 3 which correspond to 
+_R_, _G_, _B_, and _A_, respectively, of the _RGBA_ channel order.
+
+_Shape_ is the read access shape and is either AccessShape_4x1 or AccessShape_2x2. See *Access Shape*.
```
alan-baker: Access shape doesn't really add anything here since there are no descriptions associated with the enum. It might be helpful to describe somwhere what those values mean.

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+
+2+|<<_capability,Capability>>: +
+*VectorImageOperationsQCOM*
+| 7 | 4537 | <id> _Result Type_ | <<_result,'<id> Result' >> | <id> +
```
alan-baker: The anchor `_result` is not defined in this document. Should this be:

```suggestion
| 7 | 4537 | <id> _Result Type_ | <id> _Result_ | <id> +
```

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+|====
+
+[cols="1,1,7*3",width="100%"]
+|====
+7+|*OpImageVectorWriteQCOM* +
+ +
+Write multiple values into the specified image at the adjacent locations
+according to the specified access shape.
+
+_Image_ must be an object whose type is OpTypeImage.  The type must not have the Unknown image format.
```
alan-baker, line 243: You might want to require the `Sampled` operand of the image to be 0 or 2. 0 if you want to support this for OpenCL. 2 is for storage images in vulkan.

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+The behavior is undefined if the type has the image format that is not supported.
+
+
+_Coordinate_ must be a vector of <<_integer,'integer'>> type whose vector size is 2.
```
alan-baker: Similar question as for the read. Are there a limited number of supported image types or should this be based on the required coordinates for the image?

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+
+_Shape_ is the write access shape. See <<_access_shape,'Access Shape'>>.
+AccessShape_2x1 requires no additional operand. 
+AccessShape_3x1 requires one additional operand; a total of three values are written.
+AccessShape_4x1 and AccessShape_2x2 require two additional operands; a total of four values are written.
+
+_Format_ is the image format associated with _Image_. See <<_image_format,'Image Format'>>.
```
alan-baker: Is it the same format as `Image`? If so, why is this needed?

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+vectors of 32-bit / 16-bit floating point values whose lengths are 2 for NV12UVQCOM, P010UVQCOM, TP10UVQCOM,
+CompressedNV12UVQCOM, CompressedP010UVQCOM, and CompressedTP10UVQCOM
+
+32-bit floating point scalar values for UNORM_MIPI10QCOM and UNORM_MIPI12QCOM
+32-bit integer scalar values for UNSIGNED_MIPI10QCOM and UNSIGNED_MIPI12QCOM
+
+
+2+|<<_capability,Capability>>: +
+*VectorImageOperationsQCOM*
+| 6 + variable | 4538 | <id> +
```
alan-baker: I think the count is wrong here and should 7 + variable.

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
```diff
@@ -0,0 +1,313 @@
...
+_Values_ depend on the <<_image_format,image format>> associated with _Image_.
+
+32-bit / 16-bit floating point values for NV12YQCOM, P010YQCOM, TP10YQCOM, CompressedNV12YQCOM, CompressedP010YQCOM, and
+CompressedTP01YQCOM 
+
+vectors of 32-bit / 16-bit floating point values whose lengths are 2 for NV12UVQCOM, P010UVQCOM, TP10UVQCOM,
+CompressedNV12UVQCOM, CompressedP010UVQCOM, and CompressedTP10UVQCOM
+
+32-bit floating point scalar values for UNORM_MIPI10QCOM and UNORM_MIPI12QCOM
+32-bit integer scalar values for UNSIGNED_MIPI10QCOM and UNSIGNED_MIPI12QCOM
```
alan-baker: This formatting is a bit odd. Are these separate sentences or should they be bullets under `Values`?

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
wooyoungqcom: thanks. fixed

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
wooyoungqcom: you are right. thanks. fixed

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
wooyoungqcom: if I want to add a cross reference to the "Result <id>" in the SPIRV spec, what should I add instead of _result ?

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
wooyoungqcom: I made them as a bullet list. Does it make sense?

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
alan-baker: You can make it a real like using the square brackets and parentheses markup, but usually it is just left italicized.

wooyoungqcom: thanks much . fixed .

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
wooyoungqcom: A very good point. Thanks. We originally wrote it for Kernel only, then decided to make it work with Shader, too. I think we missed the part. Will update the extension.

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
AdamBrouwersHarries: The access shapes are derived from the underlying OpenCL extension that this sits upon - https://github.com/willhua/QualcommOpenCLSDKNote/blob/master/docs/extensions/cl_qcom_vector_image_ops.txt

I agree that without the context they don't add much! 

@wooyoungqcom I think re-describing the shape descriptions from the OpenCL spec may be useful. For example, it describes the `2x1` shape as: 

```
*********
* 0 * 1 *
*********
```

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
AdamBrouwersHarries: Seconded - the OpenCL environment requires that the `Sampled` operand is 0 for all images.

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
AdamBrouwersHarries: Correct - there are a specific set of image formats that are supported, and specific combinations of vector shape + image formats. As with the shapes, this is (unfortunately) only described in the underlying OpenCL specification.

Internally, we have implemented this restriction through manual checks in the SPIRV-LLVM-Translator.

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
AdamBrouwersHarries: Format is expected to be the same format as the image, as the internal mechanisms of the vector read/write functions rely on knowing the format at compile time. In the OpenCL environment we can't inspect the image for the underlying format at compile time, so this is really a workaround.  In true OpenCL fashion, programmers are expected to keep the two in sync.

AdamBrouwersHarries:
Hi @alan-baker  - thanks for your comments/feedback! As I think I mentioned in a couple of the comments, this extension was originally intended as a SPIR-V equivalent of the OpenCL extension `cl_qcom_vector_image_ops`: https://github.com/willhua/QualcommOpenCLSDKNote/blob/master/docs/extensions/cl_qcom_vector_image_ops.txt

A lot of the design should make more sense with that context, particularly the image formats, access shapes, etc. For example, we limit the image depth to 2d images, as the underlying OpenCL extension only has functions that take an `image2d_t`.

extensions/QCOM/SPV_QCOM_image_vector_read_write.asciidoc
alan-baker: I see, the image likely has an unknown format from OpenCL.

PR #450, opened 2026-09-04
build HTML specs with the sectanchors attribute

bashbaug:
see also: https://github.com/KhronosGroup/SPIRV-Docs/pull/10

It would be nice to build the extension specs in this repo with the "sectanchors" attribute, also.


alan-baker (approved):

bashbaug:
Merging as discussed in the September 23rd teleconference.

PR #451, opened 2026-09-09
Fix UTF encoding description in SPV_KHR_constant_data

### LINKED ISSUE #449: SPV_KHR_constant_data: "UTF" is ambiguous. I think you mean UTF-8
The [`SPV_KHR_constant_data`](https://github.khronos.org/SPIRV-Registry/extensions/KHR/SPV_KHR_constant_data.html) extension describes the `UTFEncodedKHR` decoration as:

> Decorate an array type with scalar integer type elements to indicate that it should be interpreted as a UTF encoded string for debugging, transpiling, or assembly/disassembly. This has no semantic impact and can be safely removed from a module. The Width of each integer element of the decorated type must be 8.

But "UTF encoded string" is ambiguous.  The Unicode standard defines three encodings: UTF-8, UTF-16, and UTF-32.

The vast majority of documents on the internet are in UTF-8.

(Amusingly, browser JavaScript by convention uses UTF-16 encodings.)

Given that the integer element in the array is required to be 8 bits wide, I think the description should be amended to say "UTF-8" instead of "UTF". (There's no need to change the name of the token):

> Decorate an array type with scalar integer type elements to indicate that it should be interpreted as a **UTF-8** encoded string for debugging, transpiling, or assembly/disassembly. This has no semantic impact and can be safely removed from a module. The Width of each integer element of the decorated type must be 8.

Tobski:
Corrected the description of the UTFEncodedKHR decoration to specify 'UTF-8' instead of 'UTF'.

Fixes https://github.com/KhronosGroup/SPIRV-Registry/issues/449

### DIFF
--- a/extensions/KHR/SPV_KHR_constant_data.asciidoc
+++ b/extensions/KHR/SPV_KHR_constant_data.asciidoc
@@ -178,7 +178,7 @@ Add this row to the Decoration table:
 2+^| Decoration ^| Extra Operands ^| Enabling Capabilities
 | 5145 | *UTFEncodedKHR* + 
 Decorate an array type with scalar _integer type_ elements to indicate that
-it should be interpreted as a UTF encoded string for debugging, transpiling,
+it should be interpreted as a UTF-8 encoded string for debugging, transpiling,
 or assembly/disassembly.
 This has no semantic impact and can be safely removed from a module.
 The _Width_ of each integer element of the decorated type must be 8.


bashbaug (approved):

bashbaug:
@dneto0 FYI.

bashbaug:
Merging as discussed in the September 23rd teleconference.

